PR #4, opened 2018-06-30
Fix directory file name

composerinteralia:
I seem to be confused about rspec-performance vs. rspec_performance.
Right now we have a mismatch between the require in the gemspec and the
version file, so the gem will not build. For now I'm going with the
dash, since it matches the name of the gem, but I'm not entirely
confident that is correct.

https://github.com/rubocop-hq/rubocop-performance/blob/c551accab8ce0710c6248163d7aeccb1819751ba/rubocop-performance.gemspec#L4

### DIFF
--- a/lib/rubocop-performance.rb
+++ b/lib/rubocop-performance.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
 require 'rubocop'
-require_relative 'rubocop_performance/version'
+require_relative 'rubocop/performance/version'
--- a/lib/rubocop/performance/version.rb
+++ b/lib/rubocop/performance/version.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Performance
+    module Version
+      STRING = '0.1.0'
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


lib/rubocop-performance.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
 require 'rubocop'
-require_relative 'rubocop_performance/version'
+require_relative 'rubocop-performance/version'
```
koic: I think that `'rubocop/performance/version'` like RuboCop RSpec will be clean as a directory structure 😃 

```ruby
require_relative 'rubocop/performance/version'
```

- https://github.com/rubocop-hq/rubocop-rspec/blob/v1.27.0/rubocop-rspec.gemspec#L2
- https://github.com/rubocop-hq/rubocop-rspec/blob/v1.27.0/lib/rubocop/rspec/version.rb

composerinteralia: 👍 I will update this PR to use that. Thanks

bbatsov: I should have noticed this in the previous PR. Nice catch, @koic!

PR #5, opened 2018-06-30
Import RuboCop performance specs

composerinteralia:
Includes:
* RSpec configurition in spec/spec_helper and .rspec,
* Rakefile with RSpec task
* .gitignore to ignore tmp files
* All performance cop specs from RuboCop

These specs are currently passing, since rubocop-performance requires
rubocop, and rubocop still has all these cops.

When I run the specs I am getting a warning from parser. At
https://github.com/whitequark/parser/blob/v2.5.1.0/lib/parser/builders/default.rb#L1625
I am getting warnings like "Unknown escape \m is ignored: /\A\m/". I
haven't tried to dig into this yet. Does it look familiar?


bbatsov:
> I am getting warnings like "Unknown escape \m is ignored: /\A\m/". I
haven't tried to dig into this yet. Does it look familiar?

Seems like invalid escape sequence in the code, don't think I've seen this so far.

PR #8, opened 2018-07-03
Create a default rake task

rrosenblum:
I was going to include RuboCop linting in this task, but RuboCop doesn't appear to be in a good state yet.


composerinteralia:
Feel free to take on adding the RuboCop setup here. And let me know if I can help.

PR #12, opened 2018-07-06
Import new spec

composerinteralia:
https://github.com/rubocop-hq/rubocop/pull/6070 added a new performance
spec. It also made some changes to a cop, but I haven't imported any cops yet.
I will get the latest rubocop master when I pull those in.


koic:
Thanks for dealing with that right away!

PR #13, opened 2018-07-06
Import performance cops

composerinteralia:
I copied over all the performance cops from RuboCop, including the
recent changes in https://github.com/rubocop-hq/rubocop/pull/6070

`rubocop/cop/performance_cops` matches `rubocop/cop/rspec_cops`: in [rubocop-rspec](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec_cops.rb)

I am open to suggestion about the call to `remove_const`. My thought was
that we would deprecate the performance cops maybe in rubocop version
0.58.0 or 0.59.0. Adding the rubocop-performance gem would be enough
to make those deprecation warnings go away. Then once rubocop 1.0.0 is
out for a bit we could do another release of rubocop-performance that
gets rid of the call to `remove_const`.

### DIFF
--- a/lib/rubocop-performance.rb
+++ b/lib/rubocop-performance.rb
@@ -2,3 +2,4 @@
 
 require 'rubocop'
 require_relative 'rubocop/performance/version'
+require_relative 'rubocop/cop/performance_cops'
--- a/lib/rubocop/cop/performance/caller.rb
+++ b/lib/rubocop/cop/performance/caller.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Performance
+      # This cop identifies places where `caller[n]`
+      # can be replaced by `caller(n..n).first`.
+      #
+      # @example
+      #   # bad
+      #   caller[1]
+      #   caller.first
+      #   caller_locations[1]
+      #   caller_locations.first
+      #
+      #   # good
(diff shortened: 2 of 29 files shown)


composerinteralia:
@koic @bbatsov I'll keep an eye on https://github.com/rubocop-hq/rubocop/tree/master/lib/rubocop/cop/performance for any changes after this gets merged. If you see changes before I do feel free to ping me and I can bring them over.

bbatsov:
> My thought was
that we would deprecate the performance cops maybe in rubocop version
0.58.0 or 0.59.0. Adding the rubocop-performance gem would be enough
to make those deprecation warnings go away. Then once rubocop 1.0.0 is
out for a bit we could do another release of rubocop-performance that
gets rid of the call to remove_const.

Yeah, that sounds like a good approach. Likely we'll just make all the performance cops in RuboCop no ops to encourage people to update faster. 

> @koic @bbatsov I'll keep an eye on https://github.com/rubocop-hq/rubocop/tree/master/lib/rubocop/cop/performance for any changes after this gets merged. If you see changes before I do feel free to ping me and I can bring them over.

Me too. I won't be accepting new performance cops until you manage to release the first version of the new gem, only bug fixes. Same goes for Rails cops until @koic is ready with the initial release of `rubocop-rails`.

Issue #11, opened 2018-07-04
Add RuboCop rake task


composerinteralia:
And run RuboCop in CI (see #10)

koic:
I'm preparing for minimal rake tasks now. I will prepare #27 and some rake tasks.

koic:
Basic rake tasks have been prepared with #27, #28, and #30. I will close the issue.

PR #15, opened 2018-07-06
Set up circleci

### LINKED ISSUE #10: Set up CI
It looks like RuboCop runs on both Travis and CI. Do we know why that is? I'm assuming we can just run on Circle using a configuration like https://github.com/rubocop-hq/rubocop-rspec/blob/master/.circleci/config.yml

We don't quite have everything set up yet (rubocop tasks, documentation tasks, etc.), but we can at least get the specs running.

In the gemspec I put the required Ruby version as `<= 2.3` (2.2 is no longer supported), so I think we can run against 2.3 and higher.

composerinteralia:
I based the circle config largely on the one from rubocop-rspec.
I removed the rubocop and documentation builds, since I haven't added
those rake tasks yet (I am working on documentation, and see #11 for
rubocop.) I also removed the Ruby 2.1 and 2.2 builds, since those
versions are no longer supported.

I went ahead and added the coverage rake task as well.

closes #10

NOTE: The tests will all fail until we get #13 in. It probably makes sense to merge that first, then rebase this and watch them pass. @bbatsov would it be possible to add this repo to circle when you have a chance?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,73 @@
+version: 2
+
+rspec: &rspec
+  steps:
+    - checkout
+    - run: bundle install
+    - run: rake spec
+
+jobs:
+  # Ruby 2.3
+  ruby-2.3-rspec:
+    docker:
+      - image: circleci/ruby:2.3
+    <<: *rspec
+
+  # Ruby 2.4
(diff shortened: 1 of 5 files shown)


bquorning:
👍

PR #21, opened 2018-08-21
Disable CaseWhenSplat and its auto-correct by default. RuboCop #6188

rrosenblum:
These are the changes being made to RuboCop in https://github.com/rubocop-hq/rubocop/pull/6188


koic:
Thank you always!

Issue #20, opened 2018-08-10
Add CC_TEST_REPORTER_ID to circle


composerinteralia:
The specs are passing, but the build is still failing with an error about this variable

koic:
#24 started to execute CI without using Code Climate. While this is not a solution to the issue, anyway it is initially aimed at the success of a test build 💦

koic:
Let's take it up again when introducing Code Climate test-reporter for CI.

PR #24, opened 2019-03-11
Update CircleCI Config

koic:
This PR updates the following.

- Use CircleCI 2.1
- CI against Ruby 2.6 and trunk (2.7-dev)
- Drop Code Climate test-reporter

This change is synchronized with CirlceCI config of RuboCop Rails
https://github.com/rubocop-hq/rubocop-rails/blob/e57ad9cb6afd3807f9b3172405bbea7105ff4bac/config/default.yml

Also this PR aims at making CI successful first.
Code Climate test-reporter pends until #20 is resolved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,73 +1,46 @@
-version: 2
-
-rspec: &rspec
-  steps:
-    - checkout
-    - run: bundle install
-    - run: rake spec
+version: 2.1
 
 jobs:
-  # Ruby 2.3
-  ruby-2.3-rspec:
+  rake_default:
+    parameters:
+      image:
+        description: "Name of the Docker image."
+        type: string
+        default: "circleci/ruby"
     docker:
-      - image: circleci/ruby:2.3
-    <<: *rspec
-
-  # Ruby 2.4
-  ruby-2.4-rspec:
-    docker:
-      - image: circleci/ruby:2.4
-    <<: *rspec
-
-  # Ruby 2.5
-  ruby-2.5-rspec:
-    docker:
-      - image: circleci/ruby:2.5
-    <<: *rspec
-
-  # JRuby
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,73 +1,46 @@
...
+        default: "circleci/ruby"
     docker:
-      - image: circleci/ruby:2.5
+      - image: << parameters.image >>
+    environment:
+      # CircleCI container has two cores, but Ruby can see 32 cores. So we
+      # configure test-queue.
+      # See https://github.com/tmm1/test-queue#environment-variables
+      TEST_QUEUE_WORKERS: 2
+      JRUBY_OPTS: "--debug -J-Xmx1000M" # get more accurate coverage data in JRuby
```
bquorning: If you don’t run coverage, you don’t need `--debug`. I think you can use `--dev` instead, which should make it faster.

koic: Thanks for your review. I updated it and I will merge this PR.

.circleci/config.yml
```diff
@@ -1,73 +1,46 @@
...
+        default: "circleci/ruby"
     docker:
-      - image: circleci/ruby:2.5
+      - image: << parameters.image >>
+    environment:
+      # CircleCI container has two cores, but Ruby can see 32 cores. So we
+      # configure test-queue.
+      # See https://github.com/tmm1/test-queue#environment-variables
+      TEST_QUEUE_WORKERS: 2
+      JRUBY_OPTS: "--dev -J-Xmx1000M" # get more accurate coverage data in JRuby
```
bquorning: “get more accurate coverage data in JRuby” doesn’t apply.

koic: Oops. I missed it 💦

PR #26, opened 2019-03-12
Synchronize config/default.yml with RuboCop Core

koic:
This PR synchronizes config/default.yml with RuboCop Core.
https://github.com/rubocop-hq/rubocop/blob/04fb08d35556b4a620e5ee612ba8b19f30892d79/config/default.yml#L1969-L2200

Some missing cop configs and metadata are copied as is.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,11 +1,27 @@
 # This is the default configuration file.
 
+Performance/Caller:
+  Description: >-
+             Use `caller(n..n)` instead of `caller`.
+  Enabled: true
+  VersionAdded: '0.49'
+
 Performance/CaseWhenSplat:
   Description: >-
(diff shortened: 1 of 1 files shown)


config/default.yml
```diff
@@ -1,35 +1,234 @@
...
+
+Performance/LstripRstrip:
+  Description: 'Use `strip` instead of `lstrip.rstrip`.'
+  Enabled: true
+  VersionAdded: '0.36'
+
 Performance/OpenStruct:
   Description: 'Use `Struct` instead of `OpenStruct`.'
   Enabled: false
+  VersionAdded: '0.61'
```
koic: 📝 In the near future, when releasing rubocop-performance gem, it is good to release it with a version number larger than this `0.61`.

PR #47, opened 2019-05-02
Fix a false negartive for `Performance/RegexpMatch`

koic:
Follow up https://github.com/rubocop-hq/rubocop/pull/6965.

This PR fixes a false negartive for `Performance/RegexpMatch` when using RuboCop 0.68 or higher.

```ruby
# example.rb
def foo
  if /re/.match(foo, 1)
    do_something
  end
end
```

## RuboCop 0.67

RuboCop 0.67 warns.

```console
% rubocop _0.67.2_ --only Performance/RegexpMatch
Inspecting 1 file
C

Offenses:

example.rb:2:6: C: Performance/RegexpMatch: Use match? instead of match
when MatchData is not used.
  if /re/.match(foo, 1)
       ^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
```

## RuboCop 0.68 (Before)

RuboCop 0.68 does not warn. This is a false negative.

```console
% rubocop _0.68.1_ --only Performance/RegexpMatch --require
rubocop-performance
Inspecting 1 file
.

1 file inspected, no offenses detected
```

## RuboCop 0.68 (After)

This PR fixes a false negative.

```console
% rubocop --only Performance/RegexpMatch --require rubocop-performance
Inspecting 1 file
C

Offenses:

example.rb:2:6: C: Performance/RegexpMatch: Use match? instead of match
when MatchData is not used.
  if /re/.match(foo, 1)
       ^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
```

This fix uses the following `<>` node pattern notation introduced in RuboCop 0.68, so it supports Ruby version 0.68 or higher.
https://github.com/rubocop-hq/rubocop/pull/6965

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#47](https://github.com/rubocop-hq/rubocop-performance/pull/47): Fix a false negartive for `Performance/RegexpMatch` when using RuboCop 0.68 or higher. ([@koic][])
+
 ## 1.1.0 (2019-04-08)
 
 ### Changes
--- a/lib/rubocop/cop/performance/regexp_match.rb
+++ b/lib/rubocop/cop/performance/regexp_match.rb
@@ -86,8 +86,8 @@ module RuboCop
 
         def_node_matcher :match_method?, <<-PATTERN
           {
+            (send _recv :match _ <int ...>)
             (send _recv :match _)
-            (send _recv :match _ (int ...))
           }
         PATTERN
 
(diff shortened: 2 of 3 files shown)


koic:
I will release RuboCop Performance 1.2.0 after being merged with this PR.

bquorning:
I think we should add a spec example that would catch this.

koic:
I'm lacking to explain. I noticed that existing tests are falling.
https://circleci.com/gh/rubocop-hq/rubocop-performance/419

So I think that there is no need to add a test to this PR because they are covered by existing tests.

I think that it is good to add a test of NodePattern to RuboCop Core if adding a test.

PR #49, opened 2019-05-04
[Fix #48] Fix RegexpMatch false positive

### LINKED ISSUE #48: False positives for `match` in RegexpMatch
The `RegexpMatch` is a bit too eager in flagging usage of `match`. Here's a simple example that is flagged when extending Rubocop's own API:

```ruby
class MyCop < Cop
  KERNEL_PATTERN = RuboCop::NodePattern.new('(const nil? :Kernel)')
  def on_send(node)
    if KERNEL_PATTERN.match(node)
      # do stuff
    end
  end
end
```

I would suggest making the `match_method?` logic more selective, and only flag usage by receiver types known to have a regexp matching `match` method.  At risk of getting too into implementation details, rather than:
```ruby
def_node_matcher :match_method?, <<-PATTERN
  {
    (send _recv :match _)
    (send _recv :match _ (int ...))
  }
PATTERN
```
`RegexpMatch` could instead do:
```ruby
def_node_matcher :match_method?, <<-PATTERN
  {
    (send {regexp str sym} :match _)
    (send {regexp str sym} :match _ (int ...))
  }
PATTERN
```
I'm happy to open a PR if this is agreeable.
--------

## Expected behavior

The snippet above should not trigger a `Performance/RegexpMatch` violation

## Actual behavior

```
Offenses:

repro.rb:4:8: C: Performance/RegexpMatch: Use match? instead of match when MatchData is not used.
    if KERNEL_PATTERN.match(node)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
```

`rubocop file.rb --only Performance/RegexpMatch` with the above snippet saved in `file.rb`

## RuboCop version

```
0.68.0 (using Parser 2.6.0.0, running on ruby 2.4.5 x86_64-darwin17)
```

dduugg:
This fixes false positives in `Performance/RegexpMatch` by only flagging single-argument `match` methods where either the receiver or the argument is a literal of type `Regexp`, `String`, `Symbol`.

Resolves https://github.com/rubocop-hq/rubocop-performance/issues/48

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#48](https://github.com/rubocop-hq/rubocop-performance/issues/48): Reduce `Performance/RegexpMatch` false positive by only flagging `match` used with Regexp/String/Symbol literals. ([@dduugg][])
+
 ### Changes
 
 * [#52](https://github.com/rubocop-hq/rubocop-performance/issues/52): Drop support for Ruby 2.2. ([@bquorning][])
@@ -10,7 +14,7 @@
 
 ### Bug fixes
 
-* [#47](https://github.com/rubocop-hq/rubocop-performance/pull/47): Fix a false negartive for `Performance/RegexpMatch` when using RuboCop 0.68 or higher. ([@koic][])
+* [#47](https://github.com/rubocop-hq/rubocop-performance/pull/47): Fix a false negative for `Performance/RegexpMatch` when using RuboCop 0.68 or higher. ([@koic][])
 
 ## 1.1.0 (2019-04-08)
 
@@ -28,3 +32,4 @@
 [@composerinteralia]: https://github.com/composerinteralia
 [@koic]: https://github.com/koic
 [@bquorning]: https://github.com/bquorning
+[@dduugg]: https://github.com/dduugg
--- a/lib/rubocop/cop/performance/regexp_match.rb
+++ b/lib/rubocop/cop/performance/regexp_match.rb
@@ -80,14 +80,14 @@ module RuboCop
         # Constants are included in this list because it is unlikely that
         # someone will store `nil` as a constant and then use it for comparison
         TYPES_IMPLEMENTING_MATCH = %i[const regexp str sym].freeze
-        MSG =
-          'Use `match?` instead of `%<current>s` when `MatchData` ' \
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,15 @@
...
+### Bug fixes
+
+* [#48](https://github.com/rubocop-hq/rubocop-performance/issues/48): Fix false positives in Performance/RegexpMatch. ([@dduugg][])
+
 ## 1.2.0 (2019-05-04)
 
 ### Bug fixes
 
-* [#47](https://github.com/rubocop-hq/rubocop-performance/pull/47): Fix a false negartive for `Performance/RegexpMatch` when using RuboCop 0.68 or higher. ([@koic][])
+* [#47](https://github.com/rubocop-hq/rubocop-performance/pull/47): Fix a false negative for `Performance/RegexpMatch` when using RuboCop 0.68 or higher. ([@koic][])
```
dduugg: unrelated typo fix

lib/rubocop/cop/performance/regexp_match.rb
```diff
@@ -80,14 +80,14 @@ class RegexpMatch < Cop
         # Constants are included in this list because it is unlikely that
         # someone will store `nil` as a constant and then use it for comparison
         TYPES_IMPLEMENTING_MATCH = %i[const regexp str sym].freeze
-        MSG =
-          'Use `match?` instead of `%<current>s` when `MatchData` ' \
+        MSG = 'Use `match?` instead of `%<current>s` when `MatchData` ' \
```
dduugg: This is to resolve a `Metrics/ClassLength` cop violation.

CHANGELOG.md
```diff
@@ -2,11 +2,15 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#48](https://github.com/rubocop-hq/rubocop-performance/issues/48): Fix false positives in `Performance/RegexpMatch`. ([@dduugg][])
```
koic: Could you write a little more about what false positives have been fixed?

koic:
This code looks good. Could you please rebase using latest master branch? A warning will probably occur because Ruby 2.2 was dropped.

dduugg:
@koic I've updated the changelog to detail the change, and rebased on current master. Please take another look. Thanks!

koic:
Thanks!

PR #51, opened 2019-05-07
[Fix #50] Make EndWith/StartWith autocorrects nil-safe

### LINKED ISSUE #50: EndWith/StartWith autocorrects are not nil-safe
The `EndWith`/`StartWith` autocorrects do not contemplate `nil`. I propose modifying them to behave like `RegexpMatch` (which corrects `do_something if pattern =~ param` into `do_something if pattern&.match?(path)`

## Expected behavior

`do_something if var =~ /\A#/` should autocorrect to `do_something if var&.start_with?('#')`

Since it appears rubocop will soon [end support](https://github.com/rubocop-hq/rubocop/issues/6945) for pre-2.3 ruby, I propose only fixing this for ruby 2.3 and above (where safe navigation is available).

## Actual behavior

`do_something if var =~ /\A#/` autocorrects to `do_something if var.start_with?('#')`. The correction will now raise a `NoMethodError` when `var` is `nil`, where previously the condition would simply evaluate to `nil`.

Invoke `rubocop -a --only Performance/StartWith` on a file containing `do_something if var =~ /\A#/`, with a `.rubocop.yml` file containing:
```
require:
  - rubocop-performance
Performance/StartWith:
  AutoCorrect: true
```

## RuboCop version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler. Here's an example:

```
$ bundle exec rubocop -V
0.68.0 (using Parser 2.6.0.0, running on ruby 2.4.5 x86_64-darwin17)
```

dduugg:
Makes the autocorrects for `Performance/EndWith` and `Performance/StartWith` safe to use with `nil` receivers, by invoking the autocorrected methods (`end_with?` and `start_with?`) with safe navigation where supported (ruby versions >= 2.3).

I've also enabled autocorrect for these cops in the default config. The autocorrect remains unsafe (particularly when correcting `Symbol` receivers, which do not support `end_with?` and `start_with?`, but this is captured by the `SafeAutoCorrect` setting. I'm happy to revert if the reviewer disagrees.

Ruby versions < 2.3 are not addressed, for reasons described in the linked ticket (impending end of support for the releases in rubocop core).

I've split the tests by ruby version where it mattered. This triggered the `RSpec/RepeatedExample` cop, which i evaded by some slight tweaks to the receiver name in the examples (`s`, `str`, `string`). Again, happy to take a different approach if the reviewer would like.

Fixes https://github.com/rubocop-hq/rubocop-performance/issues/50

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,13 @@
 ### Bug fixes
 
 * [#54](https://github.com/rubocop-hq/rubocop-performance/issues/54): Fix `Performance/FixedSize` to accept const assign with some operation. ([@tejasbubane][])
+* [#50](https://github.com/rubocop-hq/rubocop-performance/issues/50): Make `Performance/EndWith` and `Performance/StartWith` autocorrects nil-safe. ([@dduugg][])
 
 ## 1.3.0 (2019-05-13)
 
 ### Bug fixes
 
-* [#48](https://github.com/rubocop-hq/rubocop-performance/issues/48): Reduce `Performance/RegexpMatch` false positive by only flagging `match` used with Regexp/String/Symbol literals. ([@dduugg][])
+* [#48](https://github.com/rubocop-hq/rubocop-performance/issues/48): Reduce `Performance/RegexpMatch` false positives by only flagging `match` used with Regexp/String/Symbol literals. ([@dduugg][])
 
 ### Changes
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -82,7 +82,6 @@ Performance/EndWith:
   # object. Switching these methods has to be done with knowledge of the types
   # of the variables which rubocop doesn't have.
   SafeAutoCorrect: false
-  AutoCorrect: false
   Enabled: true
   VersionAdded: '0.36'
   VersionChanged: '0.44'
@@ -176,7 +175,6 @@ Performance/StartWith:
   # object. Switching these methods has to be done with knowledge of the types
   # of the variables which rubocop doesn't have.
   SafeAutoCorrect: false
-  AutoCorrect: false
   Enabled: true
   VersionAdded: '0.36'
   VersionChanged: '0.44'
--- a/lib/rubocop/cop/performance/end_with.rb
+++ b/lib/rubocop/cop/performance/end_with.rb
@@ -44,7 +44,7 @@ module RuboCop
             regex_str = interpret_string_escapes(regex_str)
 
             lambda do |corrector|
-              new_source = receiver.source + '.end_with?(' +
+              new_source = receiver.source + '&.end_with?(' +
                            to_string_literal(regex_str) + ')'
               corrector.replace(node.source_range, new_source)
             end
--- a/lib/rubocop/cop/performance/start_with.rb
+++ b/lib/rubocop/cop/performance/start_with.rb
@@ -47,7 +47,7 @@ module RuboCop
             regex_str = interpret_string_escapes(regex_str)
 
             lambda do |corrector|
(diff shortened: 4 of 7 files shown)


dduugg:
As rubocop has [dropped support](https://github.com/rubocop-hq/rubocop/pull/7026) for Ruby 2.2, I'm happy to do the same in a separate PR for this repo, which will then simplify the diff here.

koic:
Sorry for the late decision and feedback. Support for Ruby 2.2 and lower drops. So the code for Ruby versions are unnecessary.

Also, I have not yet considered whether autocorrect should be enabled by default. Please wait a moment.

dduugg:
@koic Great! I've removed all version-specific code and rebased on latest master.

In case the data point is helpful, I applied these modified cops with autocorrect to a large code base. Only 2 of 127 corrections resulted in failed tests (we have high, but not perfect, test coverage). One was due to a `NoMethodError` from invocation on a `Symbol`, the other was in code that was explicitly checking for a `nil` (rather than falsy) response.

Let me know what you decide, and I'll update the PR if necessary. Thanks for your consideration.

koic:
I've investigated the history of auto-correct not safe for `Performance/StartWith` cop and `Performance/EndWith` cop.
https://github.com/rubocop-hq/rubocop/pull/3398

Unfortunately, I think that it does not seem safe auto-correct even if safe navigation operator is used. 

```diff
% cat example.rb
obj.match?(/\Aab/)

% rubocop --require rubocop-performance -a
Inspecting 1 file
C

Offenses:

example.rb:1:1: C: [Corrected] Performance/StartWith: Use String#start_with? instead of a regex match anchored to the beginning of the string.
obj.match?(/\Aab/)
^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense corrected

% git diff
diff --git a/example.rb b/example.rb
index 8d21e67..4f59843 100644
--- a/example.rb
+++ b/example.rb
@@ -1 +1 @@
-obj.match?(/\Aab/)
+obj.start_with?('ab')
```

dduugg:
@koic We are in agreement that the autocorrect remains unsafe. Thus, `SafeAutoCorrect` remains false. I have enabled `AutoCorrect` by default, to be consistent with `RegexpMatch`. The latter is similarly unsafe:

```
$ cat example.rb
if obj.match(/ab/)
  do_stuff
end

$ rubocop --require rubocop-performance -a --only Performance/RegexpMatch
Inspecting 1 file
C

Offenses:

example.rb:1:4: C: [Corrected] Performance/RegexpMatch: Use match? instead of match when MatchData is not used.
if obj.match(/ab/)
   ^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense corrected

$ git diff
diff --git a/example.rb b/example.rb
index 05b7839..aa5bcb4 100644
--- a/example.rb
+++ b/example.rb
@@ -1,3 +1,3 @@
-if obj.match(/ab/)
+if obj.match?(/ab/)
   do_stuff
 end
```

If there is a reason to maintain the discrepancy in the default `AutoCorrect` setting between `RegexpMatch` and `EndWith`/`StartWith`, could you help me understand the distinction?

bbatsov:
@koic @dduugg Might be a good idea to wrap this PR up.

dduugg:
@bbatsov agreed, and thanks for the poke.

I'll close this out at the end of the week if the above question re: autocorrect discrepancies remains unanswered.

dduugg:
Timing out and closing, but still eager to land this if the maintainer(s) are interested.

PR #53, opened 2019-05-09
Drop support for Ruby 2.2

bquorning:
Ruby 2.2 has been EOL’ed for a while, and RuboCop just merged in a PR to drop Ruby 2.2 support. I don’t think it makes sense for RuboCop-Performance to support Ruby 2.2 now.

The 1st commit changes the Ruby version in gemspec and .rubocop.yml, and fixes the cop offenses that follows.

The 2nd commit changes all use of `strip_indent` (a utility method that is implemented in RuboCop, but might not be forever) to use “squiggly heredocs” (a Ruby 2.3 feature) instead.

-----------------

* [ ] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,7 +11,7 @@ AllCops:
     - 'vendor/**/*'
     - 'spec/fixtures/**/*'
     - 'tmp/**/*'
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.3
 
 Naming/PredicateName:
   # Method define macros for dynamically generated method.
@@ -20,9 +20,6 @@ Naming/PredicateName:
     - define_singleton_method
     - def_node_matcher
     - def_node_search
-
-Style/FrozenStringLiteralComment:
-  EnforcedStyle: always
 
 Style/FormatStringToken:
   # Because we parse a lot of source codes from strings. Percent arrays
@@ -53,9 +50,6 @@ Layout/ClassStructure:
       - instance_methods
       - protected_methods
       - private_methods
-
-Layout/IndentHeredoc:
-  EnforcedStyle: powerpack
 
 # Trailing white space is meaningful in code examples
 Layout/TrailingWhitespace:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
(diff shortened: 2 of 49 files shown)


.rubocop.yml
```diff
@@ -55,7 +55,7 @@ Layout/ClassStructure:
       - private_methods
 
 Layout/IndentHeredoc:
-  EnforcedStyle: powerpack
+  EnforcedStyle: squiggly
```
koic: Default `EnforcedStyle` of `Layout/IndentHeredoc` cop will be `squiggly`.
https://github.com/rubocop-hq/rubocop/blob/b836478a7b99786677d8377b666211a689f0573c/config/default.yml#L755-L766

So this setting can be removed.

CHANGELOG.md
```diff
@@ -21,5 +25,6 @@
 
 * Extract performance cops from rubocop-hq/rubocop repository. ([@composerinteralia][], [@koic][])
 
+[@bquorning]: https://github.com/bquorning
```
koic: Please move this link to the bottom of the changelog.

> If this is your first contribution to RuboCop project, add a link definition for the implicit link to the bottom of the changelog as [@username]: https://github.com/username.

https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format

.rubocop.yml
bquorning: The default enforced style for `Style/FrozenStringLiteralComment` will also become `always` after using the latest RuboCop. But can they be config removed already now?

bquorning: It sure seems the can. I’ll squash this change into the 1st commit.

CHANGELOG.md
bquorning: I hadn’t seen that rule, and wondered if the list was alphabetical or in date order. 😄 

:+1:

koic:
Thanks for the quick catch!

> I don’t think it makes sense for RuboCop-Performance to support Ruby 2.2 now.

I was also worried about when to drop support for Ruby 2.2. However, I think it is not necessary to maintain Ruby 2.2 only with RuboCop Performance. So, after RuboCop 0.69 is released, I'd like to release this PR (with some changes).

By the way, when will RuboCop RSpec drop Ruby 2.2?

bquorning:
> By the way, when will RuboCop RSpec drop Ruby 2.2?

I guess right after RuboCop 0.69 is released.

koic:
> I guess right after RuboCop 0.69 is released.

I get it. Thanks!

PR #58, opened 2019-05-20
Fix typo in FixedSize documentation

wteuber:


koic:
Could you run `bundle exec rake generate_cops_documentation` and squash to a single commit with the result?

wteuber:
@koic ✅ done

koic:
Thanks!

Issue #69, opened 2019-07-11
Performance Count Cop does not work


enthusiasmus:
-------

## Expected behavior

I am expecting offenses for 

```
        [1, 2, 3].select { |e| e > 2 }.size
        [1, 2, 3].reject { |e| e > 2 }.size
        [1, 2, 3].select { |e| e > 2 }.length
        [1, 2, 3].reject { |e| e > 2 }.length
```

as seen in `https://github.com/rubocop-hq/rubocop-performance/blob/master/lib/rubocop/cop/performance/count.rb`

## Actual behavior

I don't receive offenses

## Steps to reproduce the problem

file foobar.rb with 

```
# frozen_string_literal: true

def foobar
  [1, 2, 3].select { |e| e > 2 }.size
  [1, 2, 3].reject { |e| e > 2 }.size
  [1, 2, 3].select { |e| e > 2 }.length
  [1, 2, 3].reject { |e| e > 2 }.length
end
```
executing rubocop performance via `rubocop --require rubocop-performance foobar.rb`

## RuboCop version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler. Here's an example:

```
$ [bundle exec] rubocop -V
0.72.0 (using Parser 2.6.3.0, running on ruby 2.6.3 x86_64-linux)

rubocop-performance version: 1.4.0
```

koic:
Please set `SafeMode` of `Performance/Count` cop to `false` in rubocop.yml.

```yaml
Performance/Count:
  SafeMode: false
```

It is not working because `SafeMode` is `true` by default.
https://docs.rubocop.org/projects/performance/en/stable/cops_performance/#performancecount

## Related resources

- https://github.com/rubocop-hq/rubocop-performance/blob/v1.4.0/config/default.yml#L44-L48
- https://github.com/rubocop-hq/rubocop/pull/2862

rrosenblum:
I came across this in the last day or two as well. Looking into it, I think the `SafeMode` on these cops has outlived its usefulness. `SafeMode` doesn't produce a single offense for either cop so it is the same as having it disabled by default. It is only documented inside of the `default.yml` file so user's don't know that the configuration exists. I think we would be better off disabling the cop by default, or changing it to be flagged as unsafe for auto-correction.

enthusiasmus:
Thanks for the hint and reply!

koic:
#72 has been released.

PR #72, opened 2019-07-18
Remove SafeMode from Count and Detect

rrosenblum:
This is in response to #69. As noted in the issue, `SafeMode` has outlived its usefulness. Making use of being disabled by default or setting this as unsafe for auto-correction make more sense. 

I have left in the functionality that automatically disables these cops when the `Rails` cops are enabled. I'm not sure if that is functionality that we would like to maintain or not. These are the only 2 cops that function this way. If we remove the functionality, we can remove the `SafeMode` module from the core project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#74](https://github.com/rubocop-hq/rubocop-performance/pull/74): Fix an error for `Performance/RedundantMerge` when `MaxKeyValuePairs` option is set to `null`. ([@koic][])
+* [#69](https://github.com/rubocop-hq/rubocop-performance/issues/69): Remove `SafeMode` from `Performance/Count` and `Performance/Detect`. Set `SafeAutoCorrect` to `false` for these cops by default. ([@rrosenblum][])
 
 ## 1.4.1 (2019-07-29)
 
@@ -54,3 +55,4 @@
 [@bquorning]: https://github.com/bquorning
 [@dduugg]: https://github.com/dduugg
 [@tejasbubane]: https://github.com/tejasbubane
+[@rrosenblum]: https://github.com/rrosenblum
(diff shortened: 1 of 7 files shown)


koic:
> I have left in the functionality that automatically disables these cops when the Rails cops are enabled. I'm not sure if that is functionality that we would like to maintain or not. 

Yeah, I'd like to investigate whether `rails?` method can be removed. Anyway, I agree with `SafeMode` should be replaced by `SafeAutoCorrect`.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#69](https://github.com/rubocop-hq/rubocop-performance/issues/69): Remove `SafeMode` from `Performance/Count` and `Performance/Detect`. Set `SafeAutoCorrect` to `false` for these cops by default. ([@rrosenblum][])
```
koic: @rrosenblum I'd like to merge this PR. Could you move this changelog entry to new unreleased `### Changes` section?

rrosenblum:
> Yeah, I'd like to investigate whether `rails?` method can be removed.

I think there is an argument for removing it based on having `SafeAutoCorrect` set to `false` and having the reasons why it is considered unsafe being well documented.

rrosenblum:
Should I update the `VersionChanged` on these cops?

bbatsov:
Yes.

rrosenblum:
@bbatsov what are your thoughts on removing the `rails?` method in favor of `SafeAutoCorrect: false` and documentation on why it is unsafe?

bbatsov:
That sounds like a good approach to me. Let's make it happen!

rrosenblum:
Cool, I'll update this code to remove the `rails?` check. I'll then create a PR that removes the `SafeMode` module from RuboCop core.

rrosenblum:
@koic this is ready for review again.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#69](https://github.com/rubocop-hq/rubocop-performance/issues/69): Remove `SafeMode` from `Performance/Count` and `Performance/Detect`. Set `SafeAutoCorrect` to `false` for these cops by default. ([@rrosenblum][])
```
koic: Can you move `### Changes` section under `### Bug fixes` section?

rrosenblum: Updated

PR #82, opened 2019-10-27
Let Performance/StartWith and Performance/EndWith correct Regexp#match and Regexp#=~

eugeneius:
Support for `String#match?` was added in https://github.com/rubocop-hq/rubocop/pull/5744, but not `Regexp#match?`.

Calling `=~` on a regex literal generates a specialised AST node, `match-with-lvasgn`, to support assigning named captures as local variables. The pattern would previously never match this usage.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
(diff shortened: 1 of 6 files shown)


koic:
Thanks!

PR #93, opened 2020-01-09
Use `git ls-files -z` instead of splitting with newline

pocke:
It suppresses the following warning.

```bash
$ bundle exec rake 
/home/pocke/ghq/github.com/rubocop-hq/rubocop-performance/rubocop-performance.gemspec:18: warning: global variable `$RS' not initialized

(snip)
```

The gemspec will use the null character separated string for file list instead of a newline.
I think it is more appropriate because a file path can contain a newline.
And `bundle gem`'s boilerplate also uses `-z` option.


koic:
👍

PR #108, opened 2020-05-15
Fixes multi-line ReverseEach autocorrection

siegfault:
When reverse and each are on separate lines, we end up autocorrecting
```ruby
foo.reverse.
  each { ... }
```
to
```ruby
foo.reverse_
  each { ... }
```
. With this change, we will instead update to
```ruby
foo.reverse_each { ... }
```
-----------------

* [x] Added tests.

cc: @joe-sharp @dischorde

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### Bug fixes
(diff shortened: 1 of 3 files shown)


siegfault:
This has the potential to violate line length once auto corrected. Are there any concerns around that?

koic:
Yeah, I think broken code is more serious than the potential offense of `Layout/LineLength`. So, this change makes sense to me. Thank you!

siegfault:
Thanks for looking at this so quickly!

PR #117, opened 2020-05-24
Add new `Performance/StringInclude` cop

### LINKED ISSUE #81: New cop: Use String#include? instead of String#match? with /literal-only-pattern/

`String#include?` is faster than `String#match?` with a simple regexp.

https://gist.github.com/pocke/eeffeaa58ddcc35f862f6e1aca593b21
I wrote a few benchmarks, and I got the result.

The difference is small in simple cases (str1 and str2).
But there are clear differences if the string is long.

Add a cop to detect `String#match?` with simple regexp.
And I guess the cop also can detect `String#match` and `String#=~` if it is used as a condition.

## Describe alternatives you've considered

nothing

## Additional context

We already have Performance/StartWith and EndWIth. We can refer the cops to help to implement it.

fatkodima:
```ruby
# bad
'abc'.match?(/ab/)
/ab/.match?('abc')
'abc' =~ /ab/
/ab/ =~ 'abc'
'abc'.match(/ab/)
/ab/.match('abc')

# good
'abc'.include?('ab')
```

Closes #81


lib/rubocop/cop/mixin/regexp_metacharacter.rb
```diff
@@ -4,6 +4,10 @@ module RuboCop
   module Cop
     # Common functionality for handling regexp metacharacters.
     module RegexpMetacharacter
+      def literal?(regex_str)
+        regex_str =~ /\A#{Util::LITERAL_REGEX}+\z/
```
koic, line 8: Can you use `match?` instead of `=~`?

config/default.yml
```diff
@@ -196,6 +196,11 @@ Performance/StartWith:
   VersionAdded: '0.36'
   VersionChanged: '0.44'
 
+Performance/StringInclude:
+  Description: 'Use `String#include?` instead of a regex match with literal-only pattern.'
+  Enabled: true
+  VersionAdded: '1.6'
```
koic, line 202: Can you update from `'1.6'` to to `'1.7'` because patch version does not add a new cop?

fatkodima:
Updated.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#117](https://github.com/rubocop-hq/rubocop-performance/pull/117): Add new `Performance/StringInclude` cop. ([@fatkodima][])
```
koic, line 7:
```suggestion
* [#81](https://github.com/rubocop-hq/rubocop-performance/issues/81): Add new `Performance/StringInclude` cop. ([@fatkodima][])
```

lib/rubocop/cop/mixin/regexp_metacharacter.rb
```diff
@@ -6,6 +6,10 @@ module Cop
     module RegexpMetacharacter
       private
 
+      def literal?(regex_str)
+        regex_str.match?(/\A#{Util::LITERAL_REGEX}+\z/)
+      end
```
koic, line 11: It may be moved to `Performance/StringInclude` if this method is not used for other than `Performance/StringInclude`.

fatkodima:
Updated.

koic:
Thanks!

PR #119, opened 2020-05-25
[Fix #118] Fix a false positive for `DeletePrefix` and `DeletePrefix` cops

### LINKED ISSUE #118: False positive with DeletePrefix
For a multi-line string, I am using `gsub(/^  /, '')` and I receive a wrong suggestion that I should replace this call with `delete_prefix`.

Note that `^` matches at the beginning of a line, not only at the beginning of the string.
 
--------

## Expected behavior

No offense being reported.

## Actual behavior

```
lib/colorls/flags.rb:169:23: C: Performance/DeletePrefix: Use delete_prefix instead of gsub.

      puts <<EXAMPLES.gsub(/^  /, '')

                      ^^^^
```

```
def show_examples
      puts <<EXAMPLES.gsub(/^  /, '')

  examples:

    * show the given file:

      colorls README.md

    * show matching files and list matching directories:

EXAMPLES
    end
```

## RuboCop version

```
0.84.0 (using Parser 2.7.1.2, rubocop-ast 0.0.3, running on ruby 2.6.6 x86_64-linux)
```

koic:
Fixes #118.

Fix a false positive for `Performance/DeletePrefix`, `Performance/DeletePrefix`, `Performance/StartWith`, and `Performance/EndWith` cops when receiver is multiline string.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bug fixes
 
 * [#111](https://github.com/rubocop-hq/rubocop-performance/issues/111): Fix an error for `Performance/DeletePrefix` and `Performance/DeleteSuffix` cops when using autocorrection with RuboCop 0.81 or lower. ([@koic][])
+* [#118](https://github.com/rubocop-hq/rubocop-performance/issues/118): Fix a false positive for `Performance/DeletePrefix`, `Performance/DeleteSuffix`, `Performance/StartWith`, and `Performance/EndWith` cops when receiver is multiline string. ([@koic][])
 
 ## 1.6.0 (2020-05-22)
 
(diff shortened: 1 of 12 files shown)


avdv:
Hi @koic 

I think the problem is not in any way related to heredocs.

Take any string containing newlines, and you cannot simply replace a call to `.gsub(/^prefix/, '')` with `delete_prefix('prefix')` since the latter has a different meaning, likewise for `.gsub(/suffix$/`, '')`.

The only thing that works in general is `/\Aprefix/` and `/suffix\z/`.

koic:
@avdv Thanks for your feedback. I have rewritten this PR and added `SafeMultiline` option which is `true` by default. So it will detect only `/\Aprefix/` and `/suffix\z/` by default.

avdv:
Thank you, LGTM!

PR #123, opened 2020-06-02
Add new `Performance/AncestorsInclude` cop

fatkodima:
Related link - https://github.com/JuanitoFatas/fast-ruby#ancestorsinclude-vs--code

### Benchmark

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  gem 'benchmark-ips'
  gem 'benchmark-memory'
end

def fast
  (Class <= Class)
  (Class <= Module)
  (Class <= Object)
  (Class <= Kernel)
  (Class <= BasicObject)
end

def slow
  Class.ancestors.include?(Class)
  Class.ancestors.include?(Module)
  Class.ancestors.include?(Object)
  Class.ancestors.include?(Kernel)
  Class.ancestors.include?(BasicObject)
end

puts('********* IPS *********')

Benchmark.ips do |x|
  x.report('less than or equal') { fast }
  x.report('ancestors.include?') { slow }
  x.compare!
end

puts "********* MEMORY *********"

Benchmark.memory do |x|
  x.report('less than or equal') { fast }
  x.report('ancestors.include?') { slow }
  x.compare!
end
```

### Results

```
********* IPS *********
  less than or equal   444.389k i/100ms
  ancestors.include?    67.737k i/100ms
  less than or equal      4.947M (± 5.9%) i/s -     24.886M in   5.049943s
  ancestors.include?    641.527k (±27.5%) i/s -      2.709M in   5.082782s

Comparison:
  less than or equal:  4947152.6 i/s
  ancestors.include?:   641526.9 i/s - 7.71x  (± 0.00) slower

********* MEMORY *********
  less than or equal     0.000  memsize (     0.000  retained)
                         0.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
  ancestors.include?     1.000k memsize (     0.000  retained)
                         5.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)

Comparison:
  less than or equal:          0 allocated
  ancestors.include?:       1000 allocated - Infx more
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#123](https://github.com/rubocop-hq/rubocop-performance/pull/123): Add new `Performance/AncestorsInclude` cop. ([@fatkodima][])
 * [#125](https://github.com/rubocop-hq/rubocop-performance/pull/125): Support `Range#member?` method for `Performance/RangeInclude` cop. ([@fatkodima][])
 
 ## 1.6.1 (2020-06-05)
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,5 +1,11 @@
 # This is the default configuration file.
 
+Performance/AncestorsInclude:
+  Description: 'Use `A <= B` instead of `A.ancestors.include?(B)`.'
+  Reference: 'https://github.com/JuanitoFatas/fast-ruby#ancestorsinclude-vs--code'
+  Enabled: true
+  VersionAdded: '1.7'
+
 Performance/BindCall:
   Description: 'Use `bind_call(obj, args, ...)` instead of `bind(obj).call(args, ...)`.'
   Enabled: true
--- a/lib/rubocop/cop/performance/ancestors_include.rb
+++ b/lib/rubocop/cop/performance/ancestors_include.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Performance
+      # This cop is used to identify usages of `ancestors.include?` and
+      # change them to use `<=` instead.
+      #
+      # @example
+      #   # bad
+      #   A.ancestors.include?(B)
+      #
+      #   # good
+      #   A <= B
+      #
+      class AncestorsInclude < Cop
+        include RangeHelp
+
+        MSG = 'Use `<=` instead of `ancestors.include?`.'
+
+        def_node_matcher :ancestors_include_candidate?, <<~PATTERN
+          (send (send $_subclass :ancestors) :include? $_superclass)
+        PATTERN
+
+        def on_send(node)
+          return unless ancestors_include_candidate?(node)
+
+          location_of_ancestors = node.children[0].loc.selector.begin_pos
+          end_location = node.loc.selector.end_pos
+          range = range_between(location_of_ancestors, end_location)
+
+          add_offense(node, location: range)
+        end
+
+        def autocorrect(node)
+          ancestors_include_candidate?(node) do |subclass, superclass|
+            lambda do |corrector|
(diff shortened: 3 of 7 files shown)


config/default.yml
```diff
@@ -205,6 +205,11 @@ Performance/StringReplacement:
   Enabled: true
   VersionAdded: '0.33'
 
+Performance/SubclassCheck:
+  Description: 'Use `A <= B` instead of `A.ancestors.include?(B)`.'
```
koic, line 209: Can you add `Reference`?
c.f. https://github.com/rubocop-hq/rubocop-performance/blob/v1.6.0/config/default.yml#L27

lib/rubocop/cop/performance/subclass_check.rb
```diff
@@ -0,0 +1,45 @@
...
+      # change them to use `<=` instead.
+      #
+      # @example
+      #   # bad
+      #   A.ancestors.include?(B)
+      #
+      #   # good
+      #   A <= B
+      #
+      class SubclassCheck < Cop
```
koic, line 16: I think `AncestorsInclude` is a better cop name.

fatkodima:
Updated.

lib/rubocop/cop/performance_cops.rb
```diff
@@ -26,6 +26,7 @@
 require_relative 'performance/size'
 require_relative 'performance/start_with'
 require_relative 'performance/string_replacement'
+require_relative 'performance/ancestors_include'
```
koic, line 29: Can you move to the ASCII ordered position?

spec/rubocop/cop/performance/ancestors_include_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  it 'does not register an offense when using `<=`' do
+    expect_no_offenses(<<~RUBY)
+      Class <= Kernel
+    RUBY
+  end
+
+  it 'corrects `ancestors.include?` to `<=`' do
+    new_source = autocorrect_source('Class.ancestors.include?(Kernel)')
+    expect(new_source).to eq('Class <= Kernel')
```
koic, line 21: Can you use `expect_correction` instead of `autocorrect_source ... expect.to eq` in after L10?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New features
 
+* [#123](https://github.com/rubocop-hq/rubocop-performance/pull/123): Add new `Performance/AncestorsInclude` cop. ([@fatkodima][])
```
koic, line 7: Can you rebase with the latest master branch and move to under the `## master (unreleased)`?

lib/rubocop/cop/performance/ancestors_include.rb
```diff
@@ -0,0 +1,45 @@
...
+      #   # good
+      #   A <= B
+      #
+      class AncestorsInclude < Cop
+        include RangeHelp
+
+        MSG = 'Use `<=` instead of `ancestors.include?`.'
+
+        def_node_matcher :ancestors_include_candidate?, <<~PATTERN
+          (send (send $_child :ancestors) :include? $_parent)
```
koic, line 22: This is a nitpick feedback. Can you update the names?

```suggestion
          (send (send $_subclass :ancestors) :include? $_superclass)
```

In Ruby there is a method named `superclass`, so it may be a bit better as a term. E.g.: `Class.superclass #=> Module`

fatkodima:
@koic updated

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
...
+
```
koic, line 8: Can you remove redundant blank line?

fatkodima: Removed. Haven't seen it was added.

koic:
Thanks!

PR #124, opened 2020-06-03
Add new `Performance/Squeeze` cop

fatkodima:
This cop will detect places where `gsub`ing of multiple runs of single character by single character can be replaced by `squeeze` method.

```ruby
# bad
str.gsub(/a+/, 'a')
str.gsub(/\\+/, '\\')
str.gsub!(/ +/, ' ')

# good
str.squeeze('a')
str.squeeze('\\')
str.squeeze!(' ')
```

Reference - https://github.com/JuanitoFatas/fast-ruby#remove-extra-spaces-or-other-contiguous-characters-code


lib/rubocop/cop/mixin/regexp_metacharacter.rb
```diff
@@ -21,6 +21,10 @@ def literal_at_end?(regex_str)
         regex_str =~ /\A(?:#{Util::LITERAL_REGEX})+(\\z|\$)\z/
       end
 
+      def repeating_literal?(regex_str)
+        regex_str.match?(/\A(?:#{Util::LITERAL_REGEX})\+\z/)
+      end
```
koic, line 26: It may be moved to `Performance/Squeeze` if this method is not used for other than `Performance/Squeeze`.

fatkodima:
Updated.

koic:
Thanks!

PR #125, opened 2020-06-03
Support `Range#member?` method for `Performance/RangeInclude` cop

fatkodima:
```ruby
# bad
('a'..'z').member?('b')  # => true

# good
('a'..'z').cover?('b') # => true
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New features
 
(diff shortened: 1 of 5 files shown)


koic:
I think it makes sense because `Range#member?` is the alias for `Range#include?`. Can you add `VersionChanged: 1.7` to config/default.yml of this cop?

fatkodima:
Updated.

koic:
Thanks!

PR #127, opened 2020-06-06
Add new `Performance/IoReadlines` cop

fatkodima:
`IO` (and `File` as a subclass)`readlines` class (and instance) methods read full content into memory, which can be inefficient and not necessary. This cop helps to identify such places. Specifically, when `readlines` method is followed by one of `Enumerable` methods, we can safely replace it with `each_line`, which reads one line at a time thus saving memory.

We can autocorrect usages of instance level `#readlines` method, but cannot for class level `.readlines` method due to its signature. We can only show an offense and user should manually correct it. 

```ruby
# bad
File.readlines('testfile').each { |l| puts l }
IO.readlines('testfile', chomp: true).each { |l| puts l }

conn.readlines(10).map { |l| l.size }
file.readlines.find { |l| l.start_with?('#') }
file.readlines.each { |l| puts l }

# good
File.open('testfile', 'r').each_line { |l| puts l }
IO.open('testfile').each_line(chomp: true) { |l| puts l }

conn.each_line(10).map { |l| l.size }
file.each_line.find { |l| l.start_with?('#') }
file.each_line { |l| puts l }
```


lib/rubocop/cop/performance/readlines.rb
```diff
@@ -0,0 +1,127 @@
...
+      # can be replaced by `each_line` to avoid fully loading file content into memory.
+      #
+      # @example
+      #
+      #   # bad
+      #   File.readlines('testfile').each { |l| puts l }
+      #   IO.readlines('testfile', chomp: true).each { |l| puts l }
+      #
+      #   conn.readlines(10).map { |l| l.size }
+      #   file.readlines.find { |l| l.start_with?('#') }
```
koic, line 16: Looks the same as the good case.

lib/rubocop/cop/performance/readlines.rb
```diff
@@ -0,0 +1,127 @@
...
+      #
+      #   # good
+      #   File.open('testfile', 'r').each_line { |l| puts l }
+      #   IO.open('testfile').each_line(chomp: true) { |l| puts l }
+      #
+      #   conn.each_line(10).map { |l| l.size }
+      #   file.readlines.find { |l| l.start_with?('#') }
+      #   file.each_line { |l| puts l }
+      #
+      class Readlines < Cop
```
koic, line 27: It's probably better to have the class specified.

```suggestion
      class IoReadlines < Cop
```

fatkodima:
Updated.

config/default.yml
```diff
@@ -142,6 +142,11 @@ Performance/InefficientHashSearch:
   VersionAdded: '0.56'
   Safe: false
 
+Performance/IoReadlines:
+  Description: 'Use `IO.each_line` (`IO#each_line`) instead of `IO.readlines` (`IO#readlines`).'
+  Enabled: true
```
koic, line 147: I'm concerned about false positives when `readlines` is intentionally used as a useful shorthand. By default disabled (`Enabled: false`) seems better.

bbatsov:
Btw, are there any references we can link to cops like this one? E.g. from Fast Ruby or something like that? I think that'd be useful to the end users who are going over the docs.

fatkodima:
Updated: 
1. made cop disabled by default 
2. added reference (there are no anything related in fast-ruby repo, but found a good section in gitlab docs).

koic:
Thanks!

PR #129, opened 2020-06-07
Add new `Performance/BigDecimalWithNumericArgument` cop

fatkodima:
This is a simple one.

Related link - https://github.com/JuanitoFatas/fast-ruby/pull/172

I believe there is a better name for this cop.

It is faster to pass string argument to `BigDecimal` than numeric arguments.

```ruby
# bad
BigDecimal(1, 2)
BigDecimal(1.2, 3, exception: true)

# good
BigDecimal('1', 2)
BigDecimal('1.2', 3, exception: true)
```

### Benchmark
```ruby
require 'benchmark/ips'
require 'bigdecimal'

Benchmark.ips do |x|
  x.report("BigDecimal('1')")  { BigDecimal('1') }
  x.report("BigDecimal(1)")    { BigDecimal(1) }
  x.compare!
end

Benchmark.ips do |x|
  x.report("BigDecimal('1.0', 2)")  { BigDecimal('1.0', 2) }
  x.report("BigDecimal(1.0, 2)")    { BigDecimal(1.0, 2) }
  x.compare!
end
```

### Results
```
     BigDecimal('1')   179.524k i/100ms
       BigDecimal(1)   114.825k i/100ms
     BigDecimal('1')      1.782M (± 3.9%) i/s -      8.976M in   5.046628s
       BigDecimal(1)      1.107M (± 6.5%) i/s -      5.512M in   5.002115s

Comparison:
     BigDecimal('1'):  1781639.0 i/s
       BigDecimal(1):  1106879.7 i/s - 1.61x  (± 0.00) slower

BigDecimal('1.0', 2)   151.605k i/100ms
  BigDecimal(1.0, 2)    34.490k i/100ms
BigDecimal('1.0', 2)      1.519M (± 4.6%) i/s -      7.580M in   5.001610s
  BigDecimal(1.0, 2)    339.573k (± 5.0%) i/s -      1.724M in   5.093169s

Comparison:
BigDecimal('1.0', 2):  1519109.1 i/s
  BigDecimal(1.0, 2):   339573.1 i/s - 4.47x  (± 0.00) slower
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#129](https://github.com/rubocop-hq/rubocop-performance/pull/129): Add new `Performance/BigDecimalWithNumericArgument` cop. ([@fatkodima][])
 * [#130](https://github.com/rubocop-hq/rubocop-performance/pull/130): Add new `Performance/SortReverse` cop. ([@fatkodima][])
 * [#81](https://github.com/rubocop-hq/rubocop-performance/issues/81): Add new `Performance/StringInclude` cop. ([@fatkodima][])
 * [#123](https://github.com/rubocop-hq/rubocop-performance/pull/123): Add new `Performance/AncestorsInclude` cop. ([@fatkodima][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -6,6 +6,11 @@ Performance/AncestorsInclude:
   Enabled: true
   VersionAdded: '1.7'
 
+Performance/BigDecimalWithNumericArgument:
+  Description: 'Convert numeric argument to string before passing to BigDecimal.'
+  Enabled: true
+  VersionAdded: '1.7'
+
 Performance/BindCall:
   Description: 'Use `bind_call(obj, args, ...)` instead of `bind(obj).call(args, ...)`.'
   Enabled: true
--- a/docs/modules/ROOT/pages/cops.adoc
+++ b/docs/modules/ROOT/pages/cops.adoc
@@ -3,6 +3,7 @@
 = Department xref:cops_performance.adoc[Performance]
 
 * xref:cops_performance.adoc#performanceancestorsinclude[Performance/AncestorsInclude]
+* xref:cops_performance.adoc#performancebigdecimalwithnumericargument[Performance/BigDecimalWithNumericArgument]
 * xref:cops_performance.adoc#performancebindcall[Performance/BindCall]
 * xref:cops_performance.adoc#performancecaller[Performance/Caller]
 * xref:cops_performance.adoc#performancecasewhensplat[Performance/CaseWhenSplat]
--- a/docs/modules/ROOT/pages/cops_performance.adoc
+++ b/docs/modules/ROOT/pages/cops_performance.adoc
@@ -30,6 +30,36 @@ A <= B
 
 * https://github.com/JuanitoFatas/fast-ruby#ancestorsinclude-vs--code
 
+== Performance/BigDecimalWithNumericArgument
+
+|===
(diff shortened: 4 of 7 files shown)


spec/rubocop/cop/performance/big_decimal_string_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    expect_offense(<<~RUBY)
+      BigDecimal(1)
+      ^^^^^^^^^^^^^ Convert numeric argument to string before passing to `BigDecimal`.
+    RUBY
+  end
+
+  it 'registers an offense and corrects when using `BigDecimal` with float' do
+    expect_offense(<<~RUBY)
+      BigDecimal(1.5, 2, exception: true)
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Convert numeric argument to string before passing to `BigDecimal`.
```
koic, line 16: It would be better if the warning range could be specified.

```suggestion
      BigDecimal(1.5, 2, exception: true)
                 ^^^ Convert numeric argument to string before passing to `BigDecimal`.
```

spec/rubocop/cop/performance/big_decimal_string_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    RUBY
```
koic, line 17: Can you add a test for auto-correction?

lib/rubocop/cop/performance/big_decimal_string.rb
```diff
@@ -0,0 +1,51 @@
...
+        def on_send(node)
+          return unless big_decimal_string_candidate?(node)
+
+          add_offense(node)
+        end
+
+        def autocorrect(node)
+          lambda do |corrector|
+            replacement = build_replacement(node)
+            corrector.replace(node, replacement)
```
koic, line 36: Can you use `corrector.wrap` to wrap around the first argument with `'` ?

lib/rubocop/cop/performance/big_decimal_string.rb
```diff
@@ -0,0 +1,51 @@
...
+      #
+      #   # good
+      # BigDecimal('1', 2)
+      # BigDecimal('1.2', 3, exception: true)
+      #
+      class BigDecimalString < Cop
+        MSG = 'Convert numeric argument to string before passing to `BigDecimal`.'
+
+        def_node_matcher :big_decimal_string_candidate?, <<~PATTERN
+          (send nil? :BigDecimal !str_type? ...)
```
koic, line 24: I think it is better to check int and float to prevent false positives.

spec/rubocop/cop/performance/big_decimal_string_spec.rb
fatkodima: 👍 I mentioned that in the test name, but forged to write code for it 😄

fatkodima:
Updated.

lib/rubocop/cop/performance/big_decimal_string.rb
```diff
@@ -0,0 +1,43 @@
...
+      #
+      #   # bad
+      # BigDecimal(1, 2)
+      # BigDecimal(1.2, 3, exception: true)
+      #
+      #   # good
+      # BigDecimal('1', 2)
+      # BigDecimal('1.2', 3, exception: true)
+      #
+      class BigDecimalString < Cop
```
koic, line 20: I think `BigDecimalWithNumericArgument` is a better cop name.

fatkodima:
Updated with new name.

koic:
Great activity!

PR #141, opened 2020-06-21
Add new `Performance/RedundantStringChars` cop

### LINKED ISSUE #139: Need cop against `chars.slice` and al.
```ruby
# bad
str.chars.slice(0..2)
str.chars[0..2]
str.chars.first
str.chars.last

# good
str.slice(0..2)
str[0..2]
str[0]
str[-1]
```

Doing `chars` is `O(str.size)` when operation should be `O(range.size / 1)`

fatkodima:
Closes #139 

No need to call `#chars` and allocate an extra array when method can be directly called on string.

```ruby
# bad
str.chars[0..2]
str.chars.slice(0..2)

# good
str[0..2].chars

# bad
str.chars.first
str.chars.first(2)
str.chars.last
str.chars.last(2)

# good
str[0]
str[0...2].chars
str[-1]
str[-2..-1].chars

# bad
str.chars.take(2)
str.chars.drop(2)
str.chars.length
str.chars.size
str.chars.empty?

# good
str[0...2].chars
str[2..-1].chars
str.length
str.size
str.empty?
```

### Benchmark
```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  gem 'benchmark-ips'
  gem 'benchmark-memory'
end

STR = 'a' * 20
RANGE = 10..15


puts('********* IPS *********')

Benchmark.ips do |x|
  x.report("String#chars[...]")   { STR.chars[RANGE] }
  x.report("String#[...]")        { STR[RANGE] }
  x.compare!
end

puts "********* MEMORY *********"

Benchmark.memory do |x|
  x.report("String#chars[...]")   { STR.chars[RANGE] }
  x.report("String#[...]")        { STR[RANGE] }
  x.compare!
end
```

### Results
```
********* IPS *********
   String#chars[...]    18.063k i/100ms
        String#[...]   214.729k i/100ms
   String#chars[...]    181.304k (± 3.5%) i/s -    921.213k in   5.087585s
        String#[...]      2.141M (± 2.2%) i/s -     10.736M in   5.016910s

Comparison:
        String#[...]:  2141163.0 i/s
   String#chars[...]:   181304.1 i/s - 11.81x  (± 0.00) slower

********* MEMORY *********
   String#chars[...]   920.000  memsize (     0.000  retained)
                        23.000  objects (     0.000  retained)
                         2.000  strings (     0.000  retained)
        String#[...]    40.000  memsize (     0.000  retained)
                         1.000  objects (     0.000  retained)
                         1.000  strings (     0.000  retained)

Comparison:
        String#[...]:         40 allocated
   String#chars[...]:        920 allocated - 23.00x more
```


lib/rubocop/cop/performance/string_chars.rb
```diff
@@ -0,0 +1,110 @@
...
+      #   str.chars.last
+      #   str.chars.length
+      #   str.chars.size
+      #
+      #   # good
+      #   str[0..2]
+      #   str.slice(0..2)
+      #   str[0]
+      #   str[-1]
+      #   str.length
```
koic, line 23: I think that only compatible methods should be targeted.

```ruby
str = 'abcdefg'

puts '### incompatible'
puts '# str.chars.slice(0..2)'
p str.chars.slice(0..2)
puts '# str.slice(0..2)'
p str.slice(0..2)
puts

puts '### incompatible'
puts '# str.chars[0..2]'
p str.chars[0..2]
puts '# str[0..2]'
p str[0..2]
puts

puts '### compatible'
puts '# str.chars.first'
p str.chars.first
puts '# str[0]'
p str[0]
puts

puts '### compatible'
puts '# str.chars.last'
p str.chars.last
puts '# str[-1]'
p str[-1]
puts

puts '### compatible'
puts '# str.chars.length'
puts str.chars.length
puts '# str.length'
puts str.length
puts

puts '### compatible'
puts '# str.chars.size'
puts str.chars.size
puts '# str.size'
puts str.size
```

```console
% ruby example.rb
### incompatible
# str.chars.slice(0..2)
["a", "b", "c"]
# str.slice(0..2)
"abc"

### incompatible
# str.chars[0..2]
["a", "b", "c"]
# str[0..2]
"abc"

### compatible
# str.chars.first
"a"
# str[0]
"a"

### compatible
# str.chars.last
"g"
# str[-1]
"g"

### compatible
# str.chars.length
7
# str.length
7

### compatible
# str.chars.size
7
# str.size
7
```

lib/rubocop/cop/performance/string_chars.rb
```diff
@@ -0,0 +1,110 @@
...
+      #
+      class StringChars < Cop
```
koic, line 25: I think the name `RedundantStringChars` is better.

lib/rubocop/cop/performance/string_chars.rb
fatkodima: > I think that only compatible methods should be targeted.

Sure! I was dumb missing that `slice`ing and `[...]`ing an array actually returns an array 😄 
Thanks.

fatkodima:
Updated.

lib/rubocop/cop/performance/redundant_string_chars.rb
```diff
@@ -0,0 +1,101 @@
...
+      #   str.chars.first
+      #   str.chars.last
+      #   str.chars.length
+      #   str.chars.size
+      #   str.chars.empty?
+      #
+      #   # good
+      #   str[0]
+      #   str[-1]
+      #   str.length
```
koic, line 20: Can you add `str.size` to contrast with the bad case?

spec/rubocop/cop/performance/redundant_string_chars_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    expect_correction(<<~RUBY)
+      str.empty?
+    RUBY
+  end
+
+  it 'does not register an offense when using `str.chars.max`' do
+    expect_no_offenses(<<~RUBY)
+      str.chars.max
+    RUBY
+  end
```
koic, line 65: Can you add test cases to the following?

```ruby
'abc'.chars.first(2)
=> ["a", "b"]
'abc'.chars.last(2)
=> ["b", "c"]
```

fatkodima:
I thought about it a little more. This cop was originally created to primarily catch cases like `str.chars[0..2]` (and equivalent `str.chars.slice(0..2)`). This was originally implemented (incorrectly) and then dropped. I have added support to replace them with faster alternative, `str[0..2].chars`.

And also updated implementation to catch `#take(n)`, `#drop(n)`; `#first` and `#last` with argument.

config/default.yml
```diff
@@ -188,6 +188,11 @@ Performance/RedundantSortBlock:
   Enabled: true
   VersionAdded: '1.7'
 
+Performance/RedundantStringChars:
+  Description: 'Checks for places where `String#chars.<method_name>(args)` can be replaced by `String#<method_name>(args)`.'
```
koic, line 192: The following explanation may be concise and sufficient than using placeholders.

```suggestion
  Description: 'Checks for redundant `String#chars`.'
```

lib/rubocop/cop/performance/redundant_string_chars.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Performance
+      # This cop identifies places where `String#chars.<method_name>(args)`
+      # can be replaced by `String#<method_name>(args)`.
```
koic, line 7:
```suggestion
      # This cop checks for redundant `String#chars`.
```

fatkodima:
Updated.

koic:
Thanks!

PR #153, opened 2020-07-16
[Fix #145] Mark `Performance/StringInclude` as unsafe

### LINKED ISSUE #145: `Performance/StringInclude` autocorrect is unsafe
Hi,

New cop `Performance/StringInclude` introduced by #117 is unsafe when use with `--auto-correct`. 

1. It currently auto-corrects the usage of `match` and `=~`. But the return values between `match` or `=~` and `include?` are different. If the value is used, the behavior is altered.
2. `=~` doesn't raise an error in case of `nil =~ /foobar/`. Same, behavior is altered.

Two examples that broke in our codebase:
```ruby
search_by_street_address = params[:types] =~ /street_address/
@agenda_requested_in_config ||= method.match(/agenda/)
```

--- 

I think it should only fix `match?` and reserve `match` and `=~` in the case of `--auto-correct-all`. I don't know if you can make that distinction?

If not, maybe move this cop as unsafe? What do you think?

koic:
Fixes #145.

This PR marks `Performance/StringInclude` as `SafeAutocorrect: false` and disable autocorrect by default. The cop's offenses are not safe to auto-correct if a receiver is nil.

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -248,6 +248,8 @@ Performance/StartWith:
 Performance/StringInclude:
   Description: 'Use `String#include?` instead of a regex match with literal-only pattern.'
   Enabled: 'pending'
+  AutoCorrect: false
+  SafeAutoCorrect: false
   VersionAdded: '1.7'
 
 Performance/StringReplacement:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Changes
 
 * [#149](https://github.com/rubocop-hq/rubocop-performance/pull/149): Mark `Performance/AncestorsInclude` as unsafe. ([@eugeneius][])
+* [#145](https://github.com/rubocop-hq/rubocop-performance/issues/145): Mark `Performance/StringInclude` as `SafeAutocorrect: false` and disable autocorrect by default. ([@koic][])
 
 ## 1.7.0 (2020-07-07)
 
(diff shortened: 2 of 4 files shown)


config/default.yml
```diff
@@ -248,6 +248,8 @@ Performance/StartWith:
 Performance/StringInclude:
   Description: 'Use `String#include?` instead of a regex match with literal-only pattern.'
   Enabled: 'pending'
+  AutoCorrect: false
```
ghiculescu, line 251: isn't it okay to leave `AutoCorrect: true`, as long as `SafeAutoCorrect: false`?

my understanding is:

- `AutoCorrect: true` and `SafeAutoCorrect: true`= autocorrected by `rubocop -a` or `rubocop -A`
- `AutoCorrect: true` and `SafeAutoCorrect: false`= autocorrected by `rubocop -A` only
- `AutoCorrect: false` and `SafeAutoCorrect: false`= never autocorrected
- `AutoCorrect: false` and `SafeAutoCorrect: true`= never autocorrected

Currently this defaults to the third option. Why not the second?

marcandre: As it stands, `AutoCorrect: false` in the default file is reserved for Cops that have corrections that could be useful to some but probably not most people. An example that comes to my mind is `LineLength`.

This setting should be overriden in `.rubocop.yml` or via CLI for those that want it

PR #157, opened 2020-08-19
Extend `Performance/Detect` cop with check for `filter` method and `Performance/Count` cop with checks for `find_all` and `filter` methods

### LINKED ISSUE #63: [Proposal] Performance/SelectFirst
We have built this cop internally, for identifying uses of `.select {}.first` or `.select {}.count` that could be written in a more performant way. My question is, would rubocop accept a PR to add this to the main gem (if I added tests etc) or is it too niche?

```ruby
module RuboCop
  module Cop
    module TandaCustomCops
      # Don't call `.find_all { }.first`. Instead just use `.find { }`
      # The bad form iterates over every item in an array and runs the block on each.
      # The good form returns as soon as a match is found.
      class SelectFirst < Cop
        def_node_matcher :find_all_first_candidate?, <<-PATTERN
          (send (block $(send _ ${:find_all :select}) ...) ${:first} $...)
        PATTERN
        def_node_matcher :find_all_count_candidate?, <<-PATTERN
          (send (block $(send _ ${:find_all :select}) ...) ${:count} $...)
        PATTERN

        MSG_FIND = <<~MSG.freeze
          `.select { }.first` can be written as `.find { }`. This is faster because it returns as soon as a match is found.`
        MSG

        MSG_COUNT = <<~MSG.freeze
          `.select { }.count` can be written as `.count { }`. This is marginally faster.`
        MSG

        def on_send(node)
          add_offense(node, message: MSG_FIND) if find_all_first_candidate?(node)
          add_offense(node, message: MSG_COUNT) if find_all_count_candidate?(node)
        end
      end
    end
  end
end
```

fatkodima:
Closes #63 

I have rechecked existing `Detect` and `Count` cops - they are handling almost everything that is needed/requested (as mentioned in https://github.com/rubocop-hq/rubocop-performance/issues/63#issuecomment-512884529). 

So implemented checks for 2 missing methods in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Changes
 
+* [#157](https://github.com/rubocop-hq/rubocop-performance/pull/157): Extend `Performance/Detect` cop with check for `filter` method and `Performance/Count` cop with checks for `find_all` and `filter` methods. ([@fatkodima][])
 * [#154](https://github.com/rubocop-hq/rubocop-performance/pull/154): Require RuboCop 0.87 or higher. ([@koic][])
 
 ## 1.7.1 (2020-07-18)
--- a/config/default.yml
+++ b/config/default.yml
@@ -63,16 +63,14 @@ Performance/CompareWithBlock:
 
 Performance/Count:
   Description: >-
-                  Use `count` instead of `select...size`, `reject...size`,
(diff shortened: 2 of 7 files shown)


koic:
`filter` method has been added in Ruby 2.6, it may be necessary to add a version condition, but let's try early.

PR #166, opened 2020-09-05
[Fix #165] Fix a false positive for `Performance/Sum`

### LINKED ISSUE #165: Autocorrect for Performance/Sum drops initial value if it is not a number literal
When autocorrecting, the new Performance/Sum cop should take the initial value provided to `inject` or `reduce` and pass that as the first argument to `sum`, even when it is not a number literal.

This is corrected correctly:

```ruby
foo.inject(23, :+)
```

But this is not:

```ruby
foo.inject(bar, :+)
```
## Expected behavior

The code
```ruby
foo.inject(bar, :+)
```
should be corrected to 
```ruby
foo.sum(bar)
```

Also, more complex code like the following real-world example:
```ruby
paid_bills.reduce(AmountWithVAT.new, :+)
```
should be corrected to
```ruby
paid_bills.sum(AmountWithVAT.new)
```

## Actual behavior

The two examples given above are corrected to
```ruby
foo.sum
```
and
```ruby
paid_bills.sum
```

Let `rubocop-performance` autocorrect the examples given above

## RuboCop version

```
0.90.0 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.7.1 x86_64-linux)
```

`rubocop-performance` is at version 1.8.0.

koic:
Fixes #165.

This PR fixes a false positive for `Performance/Sum` when using initial value argument is a variable.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/performance/sum.rb
+++ b/lib/rubocop/cop/performance/sum.rb
@@ -98,7 +98,7 @@ module RuboCop
 
           unless init.empty?
             init = init.first
-            good_method += "(#{init.source})" if init.source.to_i != 0
+            good_method += "(#{init.source})" unless init.int_type? && init.value.zero?
           end
           good_method
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#164](https://github.com/rubocop-hq/rubocop-performance/pull/164): Fix an error for `Performance/CollectionLiteralInLoop` when a method from `Enumerable` is called with no receiver. ([@eugeneius][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/performance/sum.rb
```diff
@@ -98,7 +98,7 @@ def build_good_method(init)
 
           unless init.empty?
             init = init.first
-            good_method += "(#{init.source})" if init.source.to_i != 0
+            good_method += "(#{init.source})" unless init.int_type? && init.source.to_i.zero?
```
marcandre, line 101: `init.source.to_i.zero?` => `init.value.zero?`

koic: It's definitely simple. Thank you for your review!

marcandre (approved):

mvz:
Thanks, @koic!

koic:
@mvz You are welcome. Thank you for your feedback!

PR #173, opened 2020-09-25
Add new `Performance/BlockGivenWithExplicitBlock` cop

fatkodima:
This cop identifies unnecessary use of a `block_given?` where explicit check of block argument would suffice.

```ruby
# bad
def method(&block)
  do_something if block_given?
end

# good
def method(&block)
  do_something if block
end
```

### Benchmark
```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  gem 'benchmark-ips'
end

def if_block(&block)
  if block
  end
end

def if_block_given(&block)
  if block_given?
  end
end

Benchmark.ips do |x|
  x.report('if block')         { if_block }
  x.report('if block_given?')  { if_block_given }
  x.compare!
end
```

### Results
```
            if block     1.039M i/100ms
     if block_given?   844.838k i/100ms
            if block     10.382M (± 1.1%) i/s -     51.960M in   5.005227s
     if block_given?      8.443M (± 1.1%) i/s -     42.242M in   5.003794s

Comparison:
            if block: 10382417.4 i/s
     if block_given?:  8443073.9 i/s - 1.23x  (± 0.00) slower
```


marcandre:
Oh, cool, I had to do just that for `ostruct` and was wondering if that would trigger block capturing or not (and be slower). Glad it doesn't.

I'd mark the autocorrection as unsafe, even it's quite unlikely to be:
```ruby
def foo(&block)
  block ||= ->(x) { ...}
  warn "Using default ..." unless block_given?
  foo(&block)
end
```

fatkodima:
Good point. Updated to not trigger an offense when the block arg is reassigned.

koic (approved):

lib/rubocop/cop/performance/block_given_with_explicit_block.rb
```diff
@@ -0,0 +1,56 @@
...
+        MSG = 'Check block argument explicitly instead of using `block_given?`.'
+
+        def on_send(node)
+          def_node = node.each_ancestor(:def, :defs).first
+          return unless def_node
+
+          block_arg = def_node.arguments.find(&:blockarg_type?)
+          return unless block_arg
+
+          block_arg_name = block_arg.source[1..-1].to_sym
```
marcandre, line 40: Use `loc.name.source`; your expression isn't reliable in the unlikely scenario `def foo(&    block)`

lib/rubocop/cop/performance/block_given_with_explicit_block.rb
```diff
@@ -0,0 +1,56 @@
...
+          return if reassigns_block_arg?(def_node, block_arg_name)
+
+          add_offense(node) do |corrector|
+            corrector.replace(node, block_arg_name)
+          end
+        end
+
+        private
+
+        def reassigns_block_arg?(def_node, block_arg_name)
```
marcandre, line 50: Alternative (only if you prefer)
```ruby
def_node_matcher, :reassigns_block_arg?, '`(lvasgn %1 ...)'
```

marcandre (changes requested):

marcandre:
Sorry I forgot to check this. Don't hesitate to ping me after 2-3 days.

fatkodima:
Updated.

marcandre (approved):
LGTM 👍

koic:
Thanks!

PR #174, opened 2020-09-28
Add new `Performance/ConstantRegexp` cop

### LINKED ISSUE #151: Missing Cop for constant regexp?
The following code has 3 anti-patterns:

```ruby
def self.tokens(pattern)
   pattern.scan(TOKEN).reject { |token| token =~ /\A#{SEPARATORS}\Z/ }
end
```

This issue is to identify that `/\A#{SEPARATORS}\Z/` should be isolated as a constant, since all it's dynamic components are constants.

See: https://github.com/rubocop-hq/rubocop-ast/pull/65

fatkodima:
Closes #151 

@marcandre, mentioning you as  the original requestor :)


lib/rubocop/cop/performance/constant_regexp.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        private
+
+        def within_const_assignment?(node)
+          node.each_ancestor(:casgn).any?
+        end
+
+        def const_regexp?(node)
+          return false unless node.interpolation?
```
marcandre, line 53: `const_regexp?` need to be renamed or this check moved outside of it for clarity...

lib/rubocop/cop/performance/constant_regexp.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        MSG = 'Extract this regexp into a constant or append an `/o` option to its options.'
+
+        def on_regexp(node)
+          return if within_const_assignment?(node) ||
+                    !const_regexp?(node) ||
+                    node.single_interpolation?
+
+          add_offense(node) do |corrector|
+            corrector.replace(node, "#{node.source}o")
```
marcandre, line 42: How about `corrector.insert_after(node, 'o')`?

marcandre (commented):
Awesome!

Two nitpicks, and one additional request: could we also flag `/...#{Regexp.escape(CONST)}.../`?
I can't think of other method calls I would expect to see...

fatkodima:
Updated with suggestions.

lib/rubocop/cop/performance/constant_regexp.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        private
+
+        def within_const_assignment?(node)
+          node.each_ancestor(:casgn).any?
+        end
+
+        def_node_matcher :regexp_escape?, <<~PATTERN
+          (send
+            (const nil? :Regexp) :escape const_type?)
```
marcandre, line 54: Best use `#global_const?(:Regexp) :escape const)`

fatkodima: This should be actually written as
```
(send
  [!nil? global_const?(:Regexp)] :escape const)
```

Not sure, if this is better?

marcandre: Right, you are correct.

marcandre:
Could you please rebase? We need this :-)

fatkodima:
Rebased.

docs/modules/ROOT/pages/cops_performance.adoc
```diff
@@ -405,6 +405,46 @@ array.min_by(&:foo)
...
+| Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+
+| Pending
+| Yes
+| Yes
+| 1.9
+| -
+|===
+
+This cop finds regular expressions whose all dynamic components are constants.
```
marcandre, line 420: Just noticed this... Better description: " This cop finds regular expressions with dynamic components that are all constants"

fatkodima: Thanks 👍 Updated.

marcandre:
Awesome, thanks!

PR #176, opened 2020-10-02
Add new `Performance/MethodObjectAsBlock` cop

### LINKED ISSUE #136: Feature: New `Performance/MethodMethod`
Related link - https://github.com/JuanitoFatas/fast-ruby#normal-way-to-apply-method-vs-method-code

```ruby
# bad
array.map(&method(:do_something))

# good
array.map { |x| do_something(x) }
```

`&method(...)` construction can be used in many places, e.x. as a callback, passed with other arguments, like here https://github.com/rails/rails/blob/2ad2425fd3b3a58e937f97b1fa05306c4d3166ae/activesupport/lib/active_support/evented_file_update_checker.rb#L103

So we cannot do anything in this case.

But `&method` construction is mostly used (as I have seen) with `Enumerable` methods, like https://github.com/rails/rails/blob/2ad2425fd3b3a58e937f97b1fa05306c4d3166ae/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb#L133

I have found a couple of examples in `rails`, a couple of examples in `rubocop`, a lot of examples in `gitlab` codebase.

So I propose to detect such cases (with `Enumerable` methods) and correct them. The simplest correction (without dealing with number of arguments for each specific Enumerable method) is to replace something like this
```ruby
array.map(&method(:do_something))
```

with something like this
```ruby
array.map { |*args| do_something(*args) }
```

fatkodima:
Closes #136


lib/rubocop/cop/performance/method_object_as_block.rb
```diff
@@ -0,0 +1,33 @@
...
+      #   array.map(&method(:do_something))
+      #
+      #   # good
+      #   array.map { |x| do_something(x) }
+      #
+      class MethodObjectAsBlock < Base
+        MSG = 'Use block explicitly instead of passing a method object.'
+
+        def_node_matcher :method_object?, <<~PATTERN
+          (block-pass (send nil? :method sym))
```
marcandre, line 21: I'd switch `nil?` => `_`, and test it / document it.

lib/rubocop/cop/performance/method_object_as_block.rb
```diff
@@ -0,0 +1,33 @@
...
+        PATTERN
+
+        def on_block_pass(node)
+          method_object?(node) do
+            send_node = node.parent
+            add_offense(node) if send_node&.enumerable_method?
```
marcandre, line 27: Why restrict it to enumerable methods?

fatkodima: Yeah, I just realized that was my mistake. We should detect `&method` as an argument for any method.

lib/rubocop/cop/performance/method_object_as_block.rb
fatkodima: You mean, that we should also detect cases when `method` also has a receiver, like in this
```ruby
out = File.open('test.txt', 'w')
[ 1, 2, 3 ].each(&out.method(:puts))
```
?

marcandre: Yes, that should be `[ 1, 2, 3 ].each{ |x| out.puts(x) }`.

fatkodima:
Updated to handle `&method` with a receiver and non `Enumerable` methods.

koic (approved):

koic:
Thanks!

PR #183, opened 2020-10-19
Fix typo introduced by #167

nbennke:
Correct double named `find_all` to `find_all` and `filter` introduced by 382340966b7b065fe463680c73cb56ab4b0ab878.


koic:
Good catch! Can you update the source doc  to pass the CI?
https://github.com/rubocop-hq/rubocop-performance/blob/v1.8.1/lib/rubocop/cop/performance/detect.rb#L7

nbennke:
Thanks for hint, I'm new to CircleCi and its message was a bit confusing me, so I didn't know how to fix ist. Now I updated the sources and force-pushed the patch branch. All checks have passed :-)

koic:
Perfect! Thank you for your taking time!

PR #186, opened 2020-10-27
Update dependencies and improve version locks

AlexWayfer:
* Use stable `rubocop` version and improve its version lock.
* Drop explicit `rubocop-ast` dependency.
* Improve `simplecov` version lock.
  Not-stable versions (less than major 1) should be locked at patch-version,
  because every minor version can be not backward-compatible.
* Take `rubocop-rspec` from `Gemfile` to `.gemspec` and update it.
* Resolve new `rubocop-rspec` offenses.
* Add documentation about EOL Ruby versions support.

**Replace this text with a summary of the changes in your PR.
The more detailed you are, the better.**

-----------------

* [ ] Added tests.
* [ ] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop-performance/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop-performance/blob/master/CONTRIBUTING.md#changelog-entry-format).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,4 @@ gemspec
 gem 'bump', require: false
 gem 'rake'
 gem 'rspec'
-gem 'rubocop', github: 'rubocop-hq/rubocop'
-gem 'rubocop-rspec', '~> 1.29.0'
 gem 'yard', '~> 0.9'
--- a/rubocop-performance.gemspec
+++ b/rubocop-performance.gemspec
@@ -7,7 +7,14 @@ Gem::Specification.new do |s|
   s.name = 'rubocop-performance'
   s.version = RuboCop::Performance::Version::STRING
   s.platform = Gem::Platform::RUBY
+
+  # Using EOL Ruby versions is (gem, application) users choice
+  # (of course, not recommended).
+  # We will support them for one year after EOL for migration.
+  # Ruby 2.4 support will end at March 31, 2021:
+  # https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/
   s.required_ruby_version = '>= 2.4.0'
+
   s.authors = ['Bozhidar Batsov', 'Jonas Arvidsson', 'Yuji Nakayama']
   s.description = <<~DESCRIPTION
     A collection of RuboCop cops to check for performance optimizations
@@ -31,7 +38,7 @@ Gem::Specification.new do |s|
     'bug_tracker_uri' => 'https://github.com/rubocop-hq/rubocop-performance/issues'
   }
 
-  s.add_runtime_dependency('rubocop', '>= 0.90.0')
-  s.add_runtime_dependency('rubocop-ast', '>= 0.4.0')
-  s.add_development_dependency('simplecov')
+  s.add_runtime_dependency('rubocop', '~> 0.93.1')
+  s.add_development_dependency('rubocop-rspec', '~> 1.44')
+  s.add_development_dependency('simplecov', '~> 0.18.0')
 end
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,7 @@ workflows:
   build:
     jobs:
       - documentation-checks
+      # Check minimal required Ruby version in `gemspec` file before changes
       - rake_default:
           name: Ruby 2.4
           image: circleci/ruby:2.4
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -92,10 +92,6 @@ Performance/CollectionLiteralInLoop:
   Exclude:
     - 'spec/**/*.rb'
 
-# This disabling is a workaround for https://github.com/rubocop-hq/rubocop-rails/issues/374.
-RSpec/FactoryBot/CreateList:
-  Enabled: false
-
 RSpec/PredicateMatcher:
   EnforcedStyle: explicit
 
--- a/spec/rubocop/cop/performance/bind_call_spec.rb
+++ b/spec/rubocop/cop/performance/bind_call_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe RuboCop::Cop::Performance::BindCall, :config do
   # https://github.com/rubocop-hq/rubocop/pull/7605
   let(:ruby_version) { 2.7 }
 
-  context 'TargetRubyVersion <= 2.6', :ruby26 do
+  context 'when TargetRubyVersion <= 2.6', :ruby26 do
     it 'does not register an offense when using `bind(obj).call(args, ...)`' do
       expect_no_offenses(<<~RUBY)
         umethod.bind(obj).call(foo, bar)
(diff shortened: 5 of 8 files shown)


Gemfile
```diff
@@ -9,6 +9,4 @@ gemspec
 gem 'bump', require: false
 gem 'rake'
 gem 'rspec'
-gem 'rubocop', github: 'rubocop-hq/rubocop'
```
koic, line 12: Please keep it as this gem is catching up and dogfooding to edge RuboCop core.

Gemfile
```diff
@@ -9,6 +9,4 @@ gemspec
...
-gem 'rubocop-rspec', '~> 1.29.0'
```
koic, line 13: Also please update this line instead of adding it to the gemspec.

rubocop-performance.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |s|
     'bug_tracker_uri' => 'https://github.com/rubocop-hq/rubocop-performance/issues'
   }
 
-  s.add_runtime_dependency('rubocop', '>= 0.90.0')
-  s.add_runtime_dependency('rubocop-ast', '>= 0.4.0')
-  s.add_development_dependency('simplecov')
+  s.add_runtime_dependency('rubocop', '~> 0.93.1')
```
koic, line 34: I think it's unnecessary to drop supporting 0.90 it because the gem not depending any new features since RuboCop 0.91 or higher and I also prefer `>= 0.90.0` to provide an opportunity to upgrade to 1.0.

rubocop-performance.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |s|
...
+  s.add_development_dependency('rubocop-rspec', '~> 1.44')
+  s.add_development_dependency('simplecov', '~> 0.19.1')
```
koic, line 36: SimpleCov requires Ruby 2.5 or higher, but rubocop-performance supports for Ruby 2.4. So, the gem takes support for Ruby 2.4.

```console
#!/bin/bash -eo pipefail
bundle install
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby

    rubocop-performance was resolved to 1.8.1, which depends on
      ruby (>= 2.4.0)

    simplecov (~> 0.19.1) was resolved to 0.19.1, which depends on
      ruby (>= 2.5.0)

Exited with code exit status 6
CircleCI received exit code 6
```

Gemfile
AlexWayfer: For what? The current version of `rubocop-performance` can be incompatible with the current version of `rubocop` and CI can not catch this, or catch phantom errors.

AlexWayfer: Let's update version locks after new releases, maybe via bots like Dependabot or Depfu.

Gemfile
AlexWayfer: Why? What's wrong with `gemspec`?

rubocop-performance.gemspec
AlexWayfer: 1. I think `1.0` can be incompatible with `0.9x.x` versions.
2. I guess `>= 0.90.0` includes `x.0` versions, which definitely will not compatible.

rubocop-performance.gemspec
AlexWayfer: > but rubocop-performance supports for Ruby 2.4

For what, if Ruby 2.4 is no more maintained even for security releases?

https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/

Gemfile
koic: Yeah, CI isn't well prepared, but it's hard to keep up with the changes while waiting for the release. I like this option for maintenance.

Gemfile
koic: I can't find a reason to change to gemspec. Development dependencies are more convenient to write in Gemfile because develop can add `path` and` github` options.

rubocop-performance.gemspec
koic: As far as I know, no incompatibilities since RuboCop 0.90.0 have been found in the RuboCop Performance runtime.

rubocop-performance.gemspec
koic: Using EOL Ruby 2.4 is (gem, application) users choice (of course, not recommended). It will be maintained for one year after EOL for migration. This is a maintenance policy and I will update a doc later.

Gemfile
AlexWayfer: Maybe it should be commented line of code and anybody can "enable" it when they want?

The first, there is CI, as I wrote. The second, imagine a newcomer which want to contribute small changes but facing fails because of this unstable dependencies and becomes sad or even gives up.

Gemfile
AlexWayfer: Hm. Gems (sources) can be refine in `Gemfile`, yes, but for what then `development_dependency` in `gemspec`?

rubocop-performance.gemspec
AlexWayfer: > As far as I know, no incompatibilities since RuboCop 0.90.0 have been found in the RuboCop Performance runtime.

Maybe, but what about the second point? Like RuboCop version 2.0, 3.0 and further.

rubocop-performance.gemspec
AlexWayfer: > Using EOL Ruby 2.4 is (gem, application) users choice (of course, not recommended). It will be maintained for one year after EOL for migration. This is a maintenance policy and I will update a doc later.

OK, thank you. I've reverted this and added some documentation right into `gemspec` file above `required_ruby_version`.

koic:
I also maintain an 3rd party ConnectionAdapter gem of Active Record that needs to be tracked upstream Ruby on Rails ([rails/rails](https://github.com/rails/rails)) repository and I use that technique to this repository. And it works well.
https://github.com/rsim/oracle-enhanced/blob/v6.1.0.rc1/Gemfile#L15-L16

I'm sorry about close this suggestion because I won't change the current maintenance policy.

AlexWayfer:
> I also maintain a 3rd party ConnectionAdapter gem of Active Record that needs to be tracked upstream Ruby on Rails ([rails/rails](https://github.com/rails/rails)) repository and I use that technique to this repository. And it works well.
> https://github.com/rsim/oracle-enhanced/blob/v6.1.0.rc1/Gemfile#L15-L16
> 
> I'm sorry about close this suggestion because I won't change the current maintenance policy.

Because of you maintain some else 3rd party gem, you've rejected discussions about more popular (`rubocop-performance`) gem… it's weird.

I'm not against your "maintenance policy", I've just suggested changes and asked questions, and I left without answers. It's very sadly.

PR #188, opened 2020-11-05
[Fix #171] Extend auto-correction support for `Performance/Sum`

### LINKED ISSUE #171: Performance/Sum not autocorrecting injects or reduces with blocks
## Expected behavior

After enabling [`Performance/Sum`](https://github.com/rubocop-hq/rubocop-performance/pull/137), I expect that the following lines will all be autocorrected to use `sum`:

```
[1, 2, 3].inject(:+)
[1, 2, 3].reduce(10, :+)
[1, 2, 3].reduce { |acc, elem| acc + elem }
```

## Actual behavior

It only autocorrects the second line:
```
[1, 2, 3].inject(:+)
[1, 2, 3].sum(10)
[1, 2, 3].reduce { |acc, elem| acc + elem }
```

1. Create a test file with those lines.
2. Run `rubocop -A`
OR
You can replicate this bug by pulling down [this project](https://github.com/jmkoni/rubocop-sum-test), doing a `bundle install`, and running `rubocop -A` 

## RuboCop version

```
rubocop -V
0.91.0 (using Parser 2.7.1.4, rubocop-ast 0.4.2, running on ruby 2.7.0 x86_64-darwin19)
```

koic:
Fixes #171.

This PR makes `Performance/Sum` cop can be changed auto-correction scope depending on the value of `SafeAutoCorrect` in .rubocop.yml.

Its auto-correction is marked as safe by default (`SafeAutoCorrect: true`) to prevent `TypeError` in auto-correced code when initial value is not specified as shown below:

```ruby
['a', 'b'].sum # => (String can't be coerced into Integer)
```

Therefore if initial value is not specify as in that code, it will not be auto-corrected.

If user always want to enable auto-correction, user can set `SafeAutoCorrect: false`.

```yaml
Performance/Sum:
  SafeAutoCorrect: false
```

`SafeAutoCorrect` remains `true` by default, so the behavior does not change.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#151](https://github.com/rubocop-hq/rubocop-performance/issues/151): Add new `Performance/ConstantRegexp` cop. ([@fatkodima][])
 * [#175](https://github.com/rubocop-hq/rubocop-performance/pull/175): Add new `Performance/ArraySemiInfiniteRangeSlice` cop. ([@fatkodima][])
 * [#189](https://github.com/rubocop-hq/rubocop-performance/pull/189): Support auto-correction for `Performance/Caller`. ([@koic][])
+* [#171](https://github.com/rubocop-hq/rubocop-performance/issues/171): Extend auto-correction support for `Performance/Sum`. ([@koic][])
 
 ### Changes
 
--- a/docs/modules/ROOT/pages/cops_performance.adoc
+++ b/docs/modules/ROOT/pages/cops_performance.adoc
@@ -1788,15 +1788,39 @@ This cop identifies places where `gsub` can be replaced by
 This cop identifies places where custom code finding the sum of elements
 in some Enumerable object can be replaced by `Enumerable#sum` method.
 
+This cop can change auto-correction scope depending on the value of
+`SafeAutoCorrect`.
+Its auto-correction is marked as safe by default (`SafeAutoCorrect: true`)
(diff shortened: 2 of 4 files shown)


koic:
@rubocop-hq/rubocop-core User can change auto-correction behaviour by specifying from `SafeAutoCorrect: true (default)` to `SafeAutoCorrect: false` in .rubocop.yml. That may be an unprecedented (or unexpected) usage. If you have any concerns about using the `SafeAutoCorrect` option, please comment your feedback.

marcandre:
I'd like to know why you are proposing this instead of marking this cop as `SafeAutoCorrect: false`?

marcandre:
Oh, sorry, I understand that you want the safe autocorrection to be always done.

It would be better if `add_offense` accepted a parameter to say "this one is safe to auto-correct"

koic:
Thank you for your feedback. I can probably add to the doc about always safe auto-correction cases.
Is there anything I need to update about this PR's implementation? I'm sorry for my lack of understanding 💦

marcandre:
Actually, I like your implementation.

I think it we should have better support for this kind of cop. Maybe this cop could specify: `SafeAutoCorrect: both`?

koic:
Thank you for supplementing. I only saw it from one perspective, you've made me realize there's other perspectives to this 😅 
Currently only `SafeAutoCorrect: true` and `SafeAutoCorrect: false` are supported by core, so I'd like to start with this boolean toggle first.

PR #196, opened 2020-11-23
Fix incorrect replacement recommendation for `Performance/ChainArrayAllocation`

### LINKED ISSUE #185: Incorrect recommendation suggested by Performance/ChainArrayAllocation Cop
## Expected behavior

Expected RuboCop to suggest *valid* corrections.

## Actual behavior

```
C: Performance/ChainArrayAllocation: Use unchained values_at! and compact! (followed by return array if required)
instead of chaining values_at...compact.

options.values_at("alpha", "beta", "gamma").compact.length > 1
                                           ^^^^^^^^
```

## Problem

Hash instances **do not respond to `:values_at!`**

Run RuboCop against the following code:

```ruby
def foo?(options = {})
  options.values_at("alpha", "beta", "gamma").compact.length > 1
end
```

fatkodima:
Closes #185

The real change is this line - https://github.com/rubocop-hq/rubocop-performance/pull/196/files#diff-aadc09c37851b4bcdde6d9af4f194ba5bd2578117e64170d87d2174e17772a63R52

Other changes are a small refactoring rewriting a patterns with `%CONST` feature.

cc @marcandre

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#185](https://github.com/rubocop-hq/rubocop-performance/issues/185): Fix incorrect replacement recommendation for `Performance/ChainArrayAllocation`. ([@fatkodima][])
+
 ## 1.9.0 (2020-11-17)
 
 ### New features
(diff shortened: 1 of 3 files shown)


marcandre:
Nice refactoring. So the bug was simply a typo, then...

koic:
Thanks!

PR #241, opened 2021-05-05
[Fix #240] Disable `Performance/Casecmp` cop

### LINKED ISSUE #240: Retire `Performance/Casecmp` cop
This follows up on the ideas generated in https://github.com/rubocop/rubocop-performance/issues/100.

It is confusing to have this cop enabled. Even when the autocorrection is marked as unsafe, the cop still directs people to use `casecmp(...).zero?`, which is incorrect in non-ascii use cases.

This issue pairs with https://github.com/rubocop/rubocop/issues/9753, where I'm hoping we can create a Linting cop to suggest using `casecmp?`.

parkerfinch:
This cop suggests an incorrect fix when code involves non-ASCII
characters. Disabling it by default makes this risky behavior opt-in
rather than opt-out.

-----------------

* [ ] Added tests. (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#240](https://github.com/rubocop/rubocop-performance/issues/240): Retire `Performance/Casecmp` cop. ([@parkerfinch][])
+
 ## 1.11.2 (2021-05-05)
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


tejasbubane:
We should rather [mark the cop as unsafe](https://github.com/rubocop/rubocop-performance/pulls?q=is%3Apr+is%3Aclosed+unsafe) instead of totally disabling it.

koic:
The disadvantages of false positives with this cop seem too significant. I've decided to disable this cop by default, so this PR will be accepted. I've opened PR #443 to resolve the conflicts while preserving the commit author. Thank you.

PR #250, opened 2021-05-21
[Fix #249] Fix a false positive for `Performance/RedundantStringChars`

### LINKED ISSUE #249: RedundantStringChars auto-correct
```ruby
str = ''

# Existing code:
str.chars.last(2)
  # => []

# Rubocop suggested and auto-corrected to:
str[-2..-1].chars
  # => NoMethodError (undefined method `chars' for nil:NilClass)
# Not equivalent to `str.chars.last(2)` because below returns `nil`
str[-2..-1]
  # => nil
```

--------

## Expected behavior

Do not set this cop as auto-correctable as it can cause breakage

## Actual behavior

Auto-correct transforms into potentially broken code

NA

## RuboCop version

```
$ [bundle exec] rubocop -V
1.14.0 (using Parser 3.0.1.1, rubocop-ast 1.5.0, running on ruby 2.5.7 x86_64-linux)
  - rubocop-performance 1.6.1
  - rubocop-rails 2.5.2
```

koic:
Fixes #249.

This PR fixes a false positive for `Performance/RedundantStringChars` when using `str.chars.last` and `str.chars.drop`.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#247](https://github.com/rubocop/rubocop-performance/issues/247): Fix an incorrect auto-correct for `Performance/MapCompact` when using multi-line trailing dot method calls. ([@koic][])
+* [#249](https://github.com/rubocop/rubocop-performance/issues/249): Fix a false positive for `Performance/RedundantStringChars` when using `str.chars.last` and `str.chars.drop`. ([@koic][])
 
 ### Changes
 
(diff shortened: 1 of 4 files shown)


tejasbubane:
@koic This is not a `false positive` if it has better performance in normal cases (assuming performance was evaluated when this was implemented).

koic:
Thanks for your feedback. This PR ensures safety by focusing on safe methods. In other words, the cop name is redundant string `char`, so it is focusing on really redundant use cases.

PR #259, opened 2021-09-05
Add new `Performance/UnnecessaryStackframe` cop

casperisfine:
The size of the Ruby stack have a large impact on the performance
of exceptions.

Benchmark: https://gist.github.com/byroot/cb3bcadcc3701c2518d002fb8d3a4e7a.

This is because when an exception is raised, the backtrace has to be built
so the deeper the stack, the longer it takes.

So for performance reasons, it's beneficial to avoid creating unecessary
stack frames, as to minimize the performance impact on low level API which
commonly raise exceptions (e.g. various IO APIs, `File` etc).

NB: this is just a quick proof of concept for the most obvious pattern. The cop could certainly be made smarter to catch more cases (e.g. `{ |a| yield a }`), and likely an autocorrector as well.

But I'd like to gauge interest before I invest more time into it.

Also there's a small possibility for false positives, because `&block` might be a `lambda` or another callable with strict arity, so it's not impossible that `{ yield }` could be used to drop the block parameters. So not sure if it's a deal breaker or not. It can easily worked around by explicting the dropped arguments, e.g. `{ |_a| yield }`.


dvandersluis:
I think this is the same as `Style/ExplicitBlockArgument` on RuboCop?

byroot:
Indeed, I probably didn't notice it because it's disabled by default. Thanks.

PR #271, opened 2021-11-16
[Fix #204] Add `Performance/Sum` option to ignore potential false positives

### LINKED ISSUE #204: Add option to only enforce map/block correction of Perfomance/Sum

I have found the `Performance/Sum` cop useful in several cases but the false positive rate is too high for me because there are various cases in my codebase(s) that use `reduce(:+)` on objects where `sum` is not applicable. In cases where I'm already using `#sum`, however, there were several opportunities to use the block form which is simpler and faster. I would like to be able to have these detected without enabling the entire cop.

An option that just enforces using the block form of `#sum`  when it can be detected:

```ruby
# bad
[1, 2, 3].map { |elem| elem ** 2 }.sum
[1, 2, 3].collect(&:count).sum(10)

# good
[1, 2, 3].sum { |elem| elem ** 2 }
[1, 2, 3].sum(10, &:count)
```

## Describe alternatives you've considered

This may be too niche of an option but I think it would be helpful since I will likely disable this cop otherwise because the false positive rate for me is too high. There are also a couple of cases where I might be able to use the init parameter to `#sum` but there are cases I have where I do not prefer writing the code that way as it is not necessary or helpful AFAICT.

## Additional context

N/A

leoarnold:
Some codebases may contain non-numeric classes which implement a `:+` method.
If `a`, `b` and `c` are objects of such a class, then certain corrections
do not apply:

```ruby
[a, b].reduce(c, :+) # is equivalent to
[a, b].sum(c)

[a, b].map(&:to_i).sum # is equivalent to
[a, b].sum(&:to_i)

[a, b].reduce(:+) # works
[a, b].sum # raises TypeError
```

For users who wish to only register offenses where auto-correction
is unproblematic, we add the option `OnlySumOrWithInitialValue`
to the `Performance/Sum` cop.

The option is disabled by default, so standard behavior is preserved,
and can be activated by setting `OnlySumOrWithInitialValue: true`.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -312,3 +312,4 @@
 [@ghiculescu]: https://github.com/ghiculescu
 [@mfbmina]: https://github.com/mfbmina
 [@mvz]: https://github.com/mvz
+[@leoarnold]: https://github.com/leoarnold
--- a/changelog/new_add_performancesum_option_to_ignore.md
+++ b/changelog/new_add_performancesum_option_to_ignore.md
@@ -1,0 +1,1 @@
+* [#204](https://github.com/rubocop/rubocop-performance/issues/204): Add `Performance/Sum` option to ignore potential false positives. ([@leoarnold][])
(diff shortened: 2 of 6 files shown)


koic:
Thanks!

PR #272, opened 2021-11-16
Add Sequel compatibility mode to `Performance/Detect`

leoarnold:
When using the `sequel` gem, a typical SQL query may look like

```ruby
connection[:my_table].select { sum(nominal_value).as(total_nominal_value) }.first
```

where the `:select` method represents the SQL `SELECT` statement
and not `Enumerable#select`.

To avoid registering offenses for these cases, we add the option
`SequelCompatibilityMode` (off by default) to the `Performance/Detect`
cop. When activated, the construction

```ruby
x.select { ... }.first
```

does not constitute an offense if and only if the block passed
has zero arguments.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_sequel_compatibility_mode_to.md
+++ b/changelog/change_add_sequel_compatibility_mode_to.md
@@ -1,0 +1,1 @@
+* [#272](https://github.com/rubocop/rubocop-performance/pull/272): Add Sequel compatibility mode to `Performance/Detect`. ([@leoarnold][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -127,7 +127,8 @@ Performance/Detect:
   SafeAutoCorrect: false
   Enabled: true
   VersionAdded: '0.30'
-  VersionChanged: '1.8'
+  VersionChanged: '1.13'
+  SequelCompatibilityMode: false
 
 Performance/DoubleStartEndWith:
   Description: >-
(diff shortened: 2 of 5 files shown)


koic:
@leoarnold Thank you for opening the PR. However, RuboCop Performance focuses on plain old Ruby. So, RuboCop Performance has no framework-related features. For example, Rails performance features are managed by RuboCop Rails.

Can you suggest a cop to RuboCop Sequel for this detection of Sequel?
https://github.com/rubocop/rubocop-sequel

leoarnold:
@koic The proposed change is actually not framework specific. AR would benefit too:

https://github.com/rubocop/rubocop-performance/blob/8fe14a952b041985fa204383c49dc0347e1060f7/lib/rubocop/cop/performance/detect.rb#L34-L37

So the option could also be names `OrmCOmpatibilityMode`.

In general, I think there is no meaningful use case for `Enumerable#select` when the block passed does not take arguments. Code like

```ruby
# block has no arguments, returns same value every time
[1, 2, 3].select { whatever }.first
```

would be very strange and auto-correcting it to 

```ruby
[1, 2, 3].find { true }
```

makes it technically as fast as

```ruby
[1, 2, 3].first
```

but is it a good auto-correction to have if the code is meaningless and it is in conflict with common ORM code usage?

PR #297, opened 2022-07-27
Support autocorrection on `Performance/RedundantMatch` when receiver is a Regexp literal

r7kamura:
I found the following pattern is treated as an offense but not autocorrected, so tried to support it.

```ruby
# bad
something if /regexp/.match(string)

# good
something if /regexp/ =~ string
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_support_autocorrection_on_redundantmatch.md
+++ b/changelog/change_support_autocorrection_on_redundantmatch.md
@@ -1,0 +1,1 @@
+* [#297](https://github.com/rubocop/rubocop-performance/pull/297): Support autocorrection on `Performance/RedundantMatch` when receiver is a Regexp literal. ([@r7kamura][])
--- a/lib/rubocop/cop/performance/redundant_match.rb
+++ b/lib/rubocop/cop/performance/redundant_match.rb
@@ -41,20 +41,22 @@ module RuboCop
                         !(node.parent && node.parent.block_type?)
 
           add_offense(node) do |corrector|
-            autocorrect(corrector, node)
+            autocorrect(corrector, node) if autocorrectable?(node)
           end
         end
 
         private
 
         def autocorrect(corrector, node)
-          # Regexp#match can take a second argument, but this cop doesn't
-          # register an offense in that case
-          return unless node.first_argument.regexp_type?
-
           new_source = "#{node.receiver.source} =~ #{node.first_argument.source}"
 
           corrector.replace(node.source_range, new_source)
+        end
+
+        def autocorrectable?(node)
+          # Regexp#match can take a second argument, but this cop doesn't
+          # register an offense in that case
+          node.receiver.regexp_type? || node.first_argument.regexp_type?
         end
       end
     end
--- a/spec/rubocop/cop/performance/redundant_match_spec.rb
+++ b/spec/rubocop/cop/performance/redundant_match_spec.rb
@@ -111,4 +111,15 @@ RSpec.describe RuboCop::Cop::Performance::RedundantMatch, :config do
                    ^^^^^^^^^^^^^^^^^^ Use `=~` in places where the `MatchData` returned by `#match` will not be used.
     RUBY
   end
+
+  it 'registers an offense and corrects when receiver is a Regexp literal' do
+    expect_offense(<<~RUBY)
+      something if /regex/.match(str)
+                   ^^^^^^^^^^^^^^^^^^ Use `=~` in places where the `MatchData` returned by `#match` will not be used.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      something if /regex/ =~ str
+    RUBY
+  end
 end


changelog/change_support_autocorrection_on_redundantmatch.md
```diff
@@ -0,0 +1 @@
+* [#297](https://github.com/rubocop/rubocop-performance/pull/297): Support autocorrection on `Performance/RedundantMatch` when receiver is a Regexp. ([@r7kamura][])
```
koic, line 1:
```suggestion
* [#297](https://github.com/rubocop/rubocop-performance/pull/297): Support autocorrection on `Performance/RedundantMatch` when receiver is a Regexp literal. ([@r7kamura][])
```

lib/rubocop/cop/performance/redundant_match.rb
```diff
@@ -41,21 +41,23 @@ def on_send(node)
...
-            autocorrect(corrector, node)
+            autocorrect(corrector, node) if autocorrectable?(node)
           end
         end
 
         private
 
         def autocorrect(corrector, node)
-          # Regexp#match can take a second argument, but this cop doesn't
-          # register an offense in that case
```
koic, line 52: Can you leave the comment to `node.first_argument.regexp_type?`?

lib/rubocop/cop/performance/redundant_match.rb
```diff
@@ -41,21 +41,23 @@ def on_send(node)
...
-          return unless node.first_argument.regexp_type?
-
           new_source = "#{node.receiver.source} =~ #{node.first_argument.source}"
 
           corrector.replace(node.source_range, new_source)
         end
+
+        # @param [RuboCop::AST::SendNode] node
+        # @return [Boolean]
```
koic, line 57: In fact, type comments are unneeded :-)

```suggestion
```

spec/rubocop/cop/performance/redundant_match_spec.rb
```diff
@@ -111,4 +111,15 @@ def method(str)
                    ^^^^^^^^^^^^^^^^^^ Use `=~` in places where the `MatchData` returned by `#match` will not be used.
     RUBY
   end
+
+  it 'autocorrects Regexp#match to Regexp#=~' do
```
koic, line 115:
```suggestion
  it 'registers an offense when receiver is a Regexp literal' do
```

spec/rubocop/cop/performance/redundant_match_spec.rb
```diff
@@ -111,4 +111,15 @@ def method(str)
...
+    expect_offense(<<~TEXT)
+      something if /regex/.match(str)
+                   ^^^^^^^^^^^^^^^^^^ Use `=~` in places where the `MatchData` returned by `#match` will not be used.
+    TEXT
```
koic, line 119: nits:

```suggestion
    expect_offense(<<~RUBY)
      something if /regex/.match(str)
                   ^^^^^^^^^^^^^^^^^^ Use `=~` in places where the `MatchData` returned by `#match` will not be used.
    RUBY
```

spec/rubocop/cop/performance/redundant_match_spec.rb
r7kamura: I think it's a good to change about minor wording, but doesn't this completely lose the nuance of autocorrect support?
Or should we prefer that the message be ignorant as to whether autocorrect is supported or not?

spec/rubocop/cop/performance/redundant_match_spec.rb
r7kamura: I'm wondering if it's not a good idea to describe something that is not Ruby code as being "RUBY".

Modern editors (e.g. Visual Studio Code) recognize this heredoc delimiter as a language name and do syntax highlighting, but this heredoc includes `^^^... `, which is clearly not Ruby code and therefore it fails to parse

image

then syntax highlighting for the rest of code will be broken.

changelog/change_support_autocorrection_on_redundantmatch.md
r7kamura: 👍 

I'll update the PR title and the commit message later 👌

spec/rubocop/cop/performance/redundant_match_spec.rb
koic: `it 'registers an offense and corrects when receiver is a Regexp literal' do`, WDYT?

spec/rubocop/cop/performance/redundant_match_spec.rb
koic: I understand what you want to say. However, I think it's better to adapt to it because `RUBY` is used by convention.

lib/rubocop/cop/performance/redundant_match.rb
r7kamura: Roger.
Actually, I don't fully understand why this comment is necessary here. Why does this comment mention the second argument? `String#match` may also take a second argument, so does it need a similar comment?

spec/rubocop/cop/performance/redundant_match_spec.rb
r7kamura: It sounds good and much clear 👍

spec/rubocop/cop/performance/redundant_match_spec.rb
r7kamura: Okay, certainly it is more important thing.
Thanks for having the courage to point this out 👍

lib/rubocop/cop/performance/redundant_match.rb
koic: It just keeps the status quo for future maintenance. No text updates would be needed.

koic:
This looks good to me. Can you squash your commits into one?

PR #304, opened 2022-09-04
Change `Performance/StringInclude` autocorrection to use safe navigation operator

r7kamura:
As explained in its safety section, the autocorrection of this cop will significantly change the behavior of the code if the receiver is `nil`. I tried to use this cop's autocorrection on our Rails app, but this is the reason we couldn't apply autocorrect.

This problem can be avoided if the safe navigation operator is carefully used. It may be useless in some cases and is not a perfect solution, but it would be better than remaining possibility of `NoMethodError`. What do you think?

Note that there is another similar cop `Performance/StartWith`, which has the same issue with this.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_performance_string_include.md
+++ b/changelog/change_performance_string_include.md
@@ -1,0 +1,1 @@
+* [#303](https://github.com/rubocop/rubocop-performance/pull/303): Change `Performance/StringInclude` autocorrection to use safe navigation operator. ([@r7kamura][])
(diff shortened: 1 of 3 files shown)


r7kamura:
Another alternative is to use `.to_s.include?` instead of `&.include?`. This has the advantage that the return value is always Boolean.

```ruby
# bad
abc.match?(/ab/)

# good
abc.to_s.include?('ab')
```

koic:
I'm going to close this because concerned about the possible redundancy in safe navigation operator (even if `to_s`). And #332 was taken, which updating the description in the `safe` section. Thank you.

PR #307, opened 2022-10-01
Support autocorrection even if `reject` is used on `Performance/Count`

r7kamura:
Until now, autocorrect was given up when reject was used, but it is now supported.

```ruby
# bad
array.reject { |e| e > 2 }.size

# good
array.count { |e| !(e > 2) }
```

```ruby
# bad
array.reject { |e| e > 2 }.count

# good
array.count { |e| !(e > 2) }
```

```ruby
# bad
array.reject { |e| e > 2 }.length

# good
array.count { |e| !(e > 2) }
```

```ruby
# bad
array.reject(&:value).size

# good
array.count { |element| !element.value }
```

```ruby
# bad
array.reject(&block).size

# good
array.count { !block.call }
```

```ruby
# bad
array.reject {
  foo
  bar
}.length

# good
array.count {
  foo
  !(bar)
}
```

The last one looks a bit weird but I think other cops autocorrects it to some other good form. If there is a better way to negate statements, I'd like to use it.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/performance/count_spec.rb
+++ b/spec/rubocop/cop/performance/count_spec.rb
@@ -177,101 +177,166 @@ RSpec.describe RuboCop::Cop::Performance::Count, :config do
     end
   end
 
-  context 'autocorrect' do
-    context 'will correct' do
-      it 'select..size to count' do
-        expect_offense(<<~RUBY)
-          [1, 2].select { |e| e > 2 }.size
-                 ^^^^^^^^^^^^^^^^^^^^^^^^^ Use `count` instead of `select...size`.
-        RUBY
+  context 'with `select` and `size`' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        array.select { |e| e > 2 }.size
+              ^^^^^^^^^^^^^^^^^^^^^^^^^ Use `count` instead of `select...size`.
+      RUBY
 
-        expect_correction(<<~RUBY)
-          [1, 2].count { |e| e > 2 }
-        RUBY
-      end
-
-      it 'select..count without a block to count' do
-        expect_offense(<<~RUBY)
-          [1, 2].select { |e| e > 2 }.count
-                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `count` instead of `select...count`.
-        RUBY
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/performance/count_spec.rb
```diff
@@ -177,101 +177,139 @@ def count(&block)
...
+          foo
+          bar
+        }.length
+      RUBY
+
+      expect_correction(<<~RUBY)
+        array.count {
+          !(foo
+          bar)
+        }
```
koic, line 311: It looks like it needs to be auto-corrected to the following to be compatible.

```suggestion
        array.count {
          foo
          !bar
        }
```

r7kamura: I think that is the point I mentioned at the end of the description, do you have any better ideas to achieve that, and what about this thought?

>other cops autocorrects it to some other good form

By the way, you say "for compatible", what exactly is incompatible?

koic: Ah..., it doesn't contribute to compatibility. But `!` to `foo` makes no sense.

r7kamura: Perhaps the implementation that Solargraph uses to estimate the method's return type might be helpful:

- https://github.com/castwide/solargraph/blob/12fcf71755db2d570f591de773753eb73ac0680c/lib/solargraph/parser/legacy/node_methods.rb#L226-L256

If we have a similar implementation, we can see which nodes can be the return value of a block, and if we negate them, I think we can achieve most of what we want to do. Though this may be a little too heavy implementation for this Cop to handle.

koic: I think it can be achieved by inverting the last sentence of block with `!`. Anyway, let's make a better autocorrect separately.

r7kamura: I changed it so that the only last statement is negated in the case of `begin` node as a simple enough approach. I think indeed it's going to work there, even if it's not a perfect autocorrection.

koic:
Thanks!

PR #308, opened 2022-10-29
Mark `Performance/RegexpMatch` as unsafe

r7kamura:
I think that this Cop is not always safe for the same reason as described in the `@safety` section.

```
$ irb
irb(main):001:0> "foo".match?(/(o)/)
=> true
irb(main):002:0> Regexp.last_match
=> nil
irb(main):003:0> "foo" =~ /(o)/
=> 1
irb(main):004:0> Regexp.last_match
=> #<MatchData "o" 1:"o">
```

Whether it is good practice or not, it is not always a mistake, as it is sometimes written with the expectation that `Regexp.last_match` can be used there.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_mark_performanceregexpmatch_as_unsafe.md
+++ b/changelog/change_mark_performanceregexpmatch_as_unsafe.md
@@ -1,0 +1,1 @@
+* [#308](https://github.com/rubocop/rubocop-performance/pull/308): Mark `Performance/RegexpMatch` as unsafe. ([@r7kamura][])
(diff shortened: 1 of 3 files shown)


r7kamura:
Oh sorry, apparently I was mistaken. I thought `Regexp.last_match` would remain intact when escaping the local context, but in fact it is erased so there is no problem 👌 

```
$ irb
irb(main):001:1* def foo
irb(main):002:1*   "foo" =~ /(o)/
irb(main):003:0> end
=> :foo
irb(main):004:0> foo
=> 1
irb(main):005:0> Regexp.last_match
=> nil
```

PR #312, opened 2022-11-17
[Fix #311] New cop `Performance/IntersectPredicate`.

### LINKED ISSUE #311: Use `receiver.intersect?(argument)` instead of `(receiver & argument).any?`

Ruby 3.1 adds `Array#intersect?` which is faster than `(receiver & argument).any?` since the new method avoids creating an intermediate array while evaluating for common elements. Also new method returns true as soon as it finds a common element between arrays.

[I want to open PR, but need an issue for changelog.](https://github.com/rubocop/rubocop-performance/pull/312)

KirIgor:
In ruby 3.1 `Array#intersect?` has been added which is faster than `(receiver & argument).any?` since the new method avoids creating an intermediate array while evaluating for common elements. Also new method returns true as soon as it finds a common element between arrays.

I have added cop `Performance/IntersectPredicate` which identifies and corrects this cases. It can find `&` usage with `.any?`, `.present?`, `.empty?` and `.blank?` methods.

***

 * [x] The PR relates to only one subject with a clear title and description in grammatically correct, complete sentences.
 * [x] Wrote [good commit messages](https://chris.beams.io/posts/git-commit/).
 * [x] Commit message starts with [Fix #issue-number] (if the related issue exists).
 * [X] Feature branch is up-to-date with master (if not - rebase it).
 * [X] Added tests.
 * [X] Ran bundle exec rake default. It executes all tests and runs RuboCop on its own code.
 * [X] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop-performance/blob/master/changelog/) named {change_type}_{change_description}.md if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.


config/default.yml
```diff
@@ -181,6 +181,11 @@ Performance/InefficientHashSearch:
   VersionAdded: '0.56'
   Safe: false
 
+Performance/IntersectionCheck:
+  Description: 'Use `receiver.intersect?(argument)` instead of `(receiver & argument).any?`.'
+  Enabled: true
```
koic, line 186:
```suggestion
  Enabled: 'pending'
```

config/default.yml
```diff
@@ -181,6 +181,11 @@ Performance/InefficientHashSearch:
...
+  VersionAdded: '1.15'
```
koic, line 187:
```suggestion
  VersionAdded: '<<next>>'
```

lib/rubocop/cop/performance/intersection_check.rb
```diff
@@ -0,0 +1,71 @@
...
+      #   # bad
+      #   (receiver & argument).present?
+      #   (receiver & argument).any?
+      #   (receiver & argument).blank?
+      #   (receiver & argument).empty?
+      #
+      #   # good
+      #   receiver.intersect?(argument)
+      #   !receiver.intersect?(argument)
+      class IntersectionCheck < Base
```
koic, line 24: I'd like to avoid using the name "Check" as much as possible. What about names like "IntersectPredicate"?

```suggestion
      class IntersectPredicate < Base
```

lib/rubocop/cop/performance/intersection_check.rb
```diff
@@ -0,0 +1,71 @@
...
+      # @example
+      #   # bad
+      #   (receiver & argument).present?
+      #   (receiver & argument).any?
+      #   (receiver & argument).blank?
+      #   (receiver & argument).empty?
+      #
+      #   # good
+      #   receiver.intersect?(argument)
+      #   !receiver.intersect?(argument)
```
koic, line 23: bad and good examples can be made clearer.

```suggestion
      #   (receiver & argument).present?
      #   (receiver & argument).any?
      #
      #   # good
      #   receiver.intersect?(argument)
      #
      #   # bad
      #   (receiver & argument).blank?
      #   (receiver & argument).empty?
      #
      #   # good
      #   !receiver.intersect?(argument)
```

koic:
This is a benchmark that confirms the speedup:

```ruby
require 'bundler/inline'

gemfile(true) do
  gem 'benchmark-ips'
end

Benchmark.ips do |x|
  x.report('&') { ([1, 2, 3] & [2, 3, 4]).any? }
  x.report('intersect?') { [1, 2, 3].intersect?([2, 3, 4]) }

  x.compare!
end
```

```console
% ruby intersect.rb
(snip)

                   &   452.504k i/100ms
          intersect?   708.976k i/100ms
                   &      4.458M (± 1.8%) i/s -     22.625M in   5.076782s
          intersect?      6.818M (± 4.8%) i/s -     34.031M in   5.004272s

Comparison:
          intersect?:  6817812.7 i/s
                   &:  4457969.2 i/s - 1.53x  (± 0.00) slower
```

changelog/new_cop_intersection_check.md
```diff
@@ -0,0 +1 @@
+* [#311](https://github.com/rubocop/rubocop-performance/issues/311): Add new cop `Perfomance/IntersectionCheck` which replaces `(receiver & argument).any?` with `receiver.intersect?(argument)`, method `Array#intersect?` was added in ruby 3.1. ([@KirIgor][])
```
koic, line 1:
```suggestion
* [#311](https://github.com/rubocop/rubocop-performance/issues/311): Add new cop `Performance/IntersectionCheck` which replaces `(receiver & argument).any?` with `receiver.intersect?(argument)`, method `Array#intersect?` was added in ruby 3.1. ([@KirIgor][])
```

lib/rubocop/cop/performance/intersection_check.rb
```diff
@@ -0,0 +1,71 @@
...
+            (begin 
+              (send $(...) :& $(...))
+            ) $/present?|any?|blank?|empty?/
+          )
+        PATTERN
+
+        MSG = 'Use `%<negated>s%<receiver>s.intersect?(%<argument>s)` ' \
+              'instead of `(%<receiver>s & %<argument>s).%<method_name>s`.'
+        STRAIGHT_METHODS = %i[present? any?].freeze
+        NEGATED_METHODS = %i[blank? empty?].freeze
```
koic, line 41: Rails dependent `present?` and `blank?` should only be targeted when `ActiveSupportExtensionsEnabled` is enabled:
https://docs.rubocop.org/rubocop/1.39/configuration.html#enable-checking-active-support-extensions

koic:
I think this cop can be promoted from Performance department to Style department. Because it seems to improve readability as well as speed. Also, there doesn't seem to be any trade-offs that would cause false positives and unsafe autocorection.
What do you think?

In that case the proposal could open a PR to rubocop/rubocop repo.

lib/rubocop/cop/performance/intersection_check.rb
KirIgor: It's more clear, thank you.

KirIgor:
> I think this cop can be promoted from Performance department to Style department. Because it seems to improve readability as well as speed. Also, there doesn't seem to be any trade-offs that would cause false positives and unsafe autocorection. What do you think?
> 
> In that case the proposal could open a PR to rubocop/rubocop repo.

Yes, I think it is great idea. Operator `&` may be used as bitwise operator but check for array methods at the end excludes this, so I don't see false positives. I will then fix all discussions in PR to rubocop/rubocop. Thank you!

PR #319, opened 2022-12-01
Extend `Performance/StringInclude` to handle `!~`

### LINKED ISSUE #318: Add support for !~ to Performance/StringInclude
Currently, `Performance/StringInclude` recognizes `=~` but not `!~`:
```ruby
str =~ /foo/ # cop recommends str.include?('foo') ✔️
str !~ /foo/ # cop ignores 😔
```

Ideally, it would recommend `!str.include?('foo')`

fatkodima:
Closes #318.

### DIFF
--- a/changelog/change_string_include_handle_not_match.md
+++ b/changelog/change_string_include_handle_not_match.md
@@ -1,0 +1,1 @@
+* [#318](https://github.com/rubocop/rubocop-performance/issues/318): Extend `Performance/StringInclude` to handle `!~`. ([@fatkodima][])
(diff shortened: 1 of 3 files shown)


config/default.yml
```diff
@@ -325,7 +325,7 @@ Performance/StringInclude:
   Enabled: 'pending'
   SafeAutoCorrect: false
   VersionAdded: '1.7'
-  VersionChanged: '1.12'
+  VersionChanged: '1.16'
```
koic, line 328: Can you revert it? Because this update is not required.

```suggestion
  VersionChanged: '1.12'
```

fatkodima: Reverted.

PR #320, opened 2022-12-01
Fix `Performance/RegexpMatch` to handle `::Regexp`

### LINKED ISSUE #314: Performance/RegexpMatch: Does not consider `::Regexp.last_match(1)`
I noticed a couple of issues with this cop.

1. It does not recognize `::Regexp.last_match(1)` as use of `MatchData`
2. It gets confused and assumes use of `MatchData` if a following unrelated line mentions MatchData-like statements.

--------

### Expectation 1: `::Regexp.last_match(1)` should be detected as `MatchData` use

Expecting the cop detect `::Regexp.last_match(1)` as use of `MatchData` and not trigger an offense (in the same way it detects `Regexp.last_match(1)` or `$1`).

```ruby
# Expected not to fail
::Regexp.last_match(1) if string =~ /hello (.*)/
```

### Expectation 2: Offense should be reported

The first line below does not use `MatchData`, the cop is expected to report an offense. Removing the second (unrelated) line makes it work as expected.

```ruby
p 'hello' if string =~ /hello (.*)/
p $1 if string =~ /hello (.*)/
```

## Actual behavior

This code triggers an offense where it shouldn't:

```ruby
p ::Regexp.last_match(1) if string =~ /hello (.*)/
```

This code does not trigger an offense, where it should:

```ruby
p 'hello' if string =~ /hello (.*)/
p $1 if string =~ /hello (.*)/
```

```yaml
# .rubocop.yml

require: rubocop-performance

AllCops:
  NewCops: enable
  TargetRubyVersion: 3.0
```

```ruby
# test.rb

string = 'hello bob'

# FAIL: Offense is reported on the last line. The first two lines pass as
#       expected
p $1 if string =~ /hello (.*)/
p Regexp.last_match(1) if string =~ /hello (.*)/
p ::Regexp.last_match(1) if string =~ /hello (.*)/

# FAIL: Offense is NOT reported on first line since we use $1 in the
#       unrelated next line
p 'hello' if string =~ /hello (.*)/
p $1 if string =~ /hello (.*)/
```

then run `rubocop --only Performance` and get this output (trimmed for readability):

```
➜ rubocop --only Performance
test.rb:7:29: C: [Correctable] Performance/RegexpMatch: Use match? instead of =~ when MatchData is not used.
p ::Regexp.last_match(1) if string =~ /hello (.*)/
                            ^^^^^^^^^^^^^^^^^^^^^^
```

## Additional consideration

Note that normally I am using `$1` and not `Regexp.last_match` (and definitely not `::Regexp.last_match`), but the native `Style/PerlBackrefs` cop autocorrected my code to `::Regexp.last_match`.

The point being - if one cop is saying "use `::Regexp.last_match`", then the other cop should be aware of it.

## RuboCop version

```
$ rubocop -V
1.39.0 (using Parser 3.1.2.1, rubocop-ast 1.23.0, running on ruby 3.1.0) [x86_64-linux]
  - rubocop-performance 1.15.1
```

fatkodima:
Fixes #314.

### DIFF
--- a/changelog/fix_regexp_match_to_handle_cbase_regexp.md
+++ b/changelog/fix_regexp_match_to_handle_cbase_regexp.md
@@ -1,0 +1,1 @@
+* [#314](https://github.com/rubocop/rubocop-performance/issues/314): Fix `Performance/RegexpMatch` to handle `::Regexp`. ([@fatkodima][])
--- a/lib/rubocop/cop/performance/regexp_match.rb
+++ b/lib/rubocop/cop/performance/regexp_match.rb
@@ -124,8 +124,8 @@ module RuboCop
 
         def_node_search :last_matches, <<~PATTERN
           {
-            (send (const nil? :Regexp) :last_match)
-            (send (const nil? :Regexp) :last_match _)
+            (send (const {nil? cbase} :Regexp) :last_match)
+            (send (const {nil? cbase} :Regexp) :last_match _)
             ({back_ref nth_ref} _)
             (gvar #match_gvar?)
           }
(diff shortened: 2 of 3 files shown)


koic:
👍

PR #332, opened 2023-01-07
Register offenses for variables against regexes in `Performance/StringInclude`

fatkodima:
See https://github.com/rails/rails/pull/46910.

The second pattern of `StringInclude#redundant_regex?`(`(send (regexp (str $#literal?) (regopt)) {:match :match?} $str)`) is pretty useless, because it only detects cases like `'str'.match?(/str/)`, so literal regex is matched against literal string. But does anyone write such code?

That pattern was implemented that way, because, to avoid cases when the receiver is a symbol and symbol does not implement `#include?`. But, it already marks as offense and autocorrects `can_be_a_symbol.match?(/abc/)`! So I think this PR changes are fair.


koic:
Yeah, This change looks fair! Thanks!

PR #337, opened 2023-01-26
[Fix #336] Expand the safety warning on Performance/Detect

### LINKED ISSUE #336: Performance/Detect suggests reverse.find should be used on hash but reverse doesn't exist
`Hash` does not inherit Enumerable so `reverse` does not exist on it, but it does implement some methods with the same interface, such as `select` and `find_all`.

This came from a regression introduced in https://github.com/puppetlabs/puppetlabs_spec_helper/commit/24d92da1ce27de1fcd532dfaac440233429b7a0f#diff-77f5c8eb6c2374258943fce12a490c421cf882db8f8fdec3bd02e74bc76fac85L319

A simplified version of the above code is:

```ruby
x = {}
x.select { |_k, v| v.key?('thread') }.last
```

## Expected behavior

Ideally Rubocop would be able to figure out that the variable `x` is a hash and accept it as valid code.

## Actual behavior

It finds an offense and makes an invalid suggestion.
```
test.rb:2:3: C: [Correctable] Performance/Detect: Use reverse.find instead of select.last.
x.select { |_k, v| v.key?('thread') }.last
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## RuboCop version

```console
$ bundle exec rubocop -V
1.6.1 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.1.3 x86_64-linux)
  - rubocop-performance 1.9.2
  - rubocop-rspec 2.0.1
```

ekohl:
In addition to the external library activerecord, the built in Hash class also has compatibility issues.

I'm unsure if this deserves a changelog entry, given it doesn't change the code there shouldn't be an impact (which is why I didn't add tests either).

-----------------

* [ ] <del>Added tests.</del>

### DIFF
--- a/lib/rubocop/cop/performance/detect.rb
+++ b/lib/rubocop/cop/performance/detect.rb
@@ -8,9 +8,12 @@ module RuboCop
       # `detect` instead.
       #
       # @safety
-      #   This cop is unsafe because is has known compatibility issues with `ActiveRecord` and other
-      #   frameworks. `ActiveRecord` does not implement a `detect` method and `find` has its own
-      #   meaning. Correcting `ActiveRecord` methods with this cop should be considered unsafe.
+      #   This cop is unsafe because is assumes the class implements the
+      #   `Enumerable` interface, but can't reliably detect this. This creates
+      #   known compatibility issues with `Hash`, `ActiveRecord` and other
+      #   frameworks. `Hash` and `ActiveRecord` do not implement a `detect`
+      #   method and `find` has its own meaning. Correcting `Hash` and
+      #   `ActiveRecord` methods with this cop should be considered unsafe.
       #
       # @example
       #   # bad


koic:
Yeah, this change does not require a changelog entry and tests. Thank you for the improvement!

r7kamura:
I understand that this cop's autocorrection can fail depending on the receiver. However, I think that the autocorrection will succeed correctly both when the receiver is a `Hash` and a `ActiveRecord::Relation`. Can you give me an example of in which case it would fail?

ekohl:
In the original issue I added an example: https://github.com/rubocop/rubocop-performance/issues/336.

ekohl:
I think you're right that it could be phrased better. Should this be considered sufficient:

> This cop is unsafe because is assumes the class implements the `Enumerable` interface, but can't reliably detect this.

Then drop this part with specific method names?

> Hash and ActiveRecord do not implement a detect method and find has its own meaning.

So it becomes:

> This cop is unsafe because is assumes the class implements the `Enumerable` interface, but can't reliably detect this. This creates known compatibility issues with `Hash`, `ActiveRecord` and other frameworks. Correcting `Hash` and `ActiveRecord` methods with this cop should be considered unsafe.

r7kamura:
hmm, since this cop depends on `Array#reverse`, it seems more appropriate to write the following:

> This cop is unsafe because it assumes that the receiver is an `Array` or equivalent, but can't reliably detect it.

On another point, I feel it would be better to remove the explanation about ActiveRecord. Because in fact `#find` in `ActiveRecord::Relation` delegates to `Enumerable#find` when a block is passed, and `ActiveRecord::Relation` can also respond to `#reverse`, so it shouldn't be a problem.

https://github.com/rails/rails/blob/v7.0.4.2/activerecord/lib/active_record/relation/finder_methods.rb#L68

PR #340, opened 2023-02-11
Enhance `RedundantEqualityComparisonBlock` to check `=~` and `match?` methods

### LINKED ISSUE #339: Enhance `Rails/RedundantEqualityComparisonBlock` to support regular expressions
`Rails/RedundantEqualityComparisonBlock` could be enhanced to recognize blocks that perform regex matching:

```ruby
# bad
list.all? { |item| item =~ /pattern/ }
list.all? { |item| /pattern/.match? item }

# good
list.all?(/pattern/)
```

fatkodima:
Closes #339.

`Regexp#match` technically can be added too, but this can produce more false positives due to its name popularity 🤔


koic:
> `Regexp#match` technically can be added too, but this can produce more false positives due to its name popularity

I think not including it is a good choice. Thanks!

PR #342, opened 2023-02-21
Add new `Performance/PredicateOnSelectResult` cop

fatkodima:
Identifies usages of `any?`, `empty?` or `none?` predicate methods chained to `select`/`filter` and change them to use predicate method instead.

```ruby
# bad
arr.select { |x| x > 1 }.any?

# good
arr.any? { |x| x > 1 }

# bad
arr.select { |x| x > 1 }.empty?
arr.select { |x| x > 1 }.none?

# good
arr.none? { |x| x > 1 }

# good
relation.select(:name).any?
arr.select { |x| x > 1 }.any?(&:odd?)
```

### Benchmarks
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
  gem "benchmark-ips"
end

arr = [1,2,3]

Benchmark.ips do |x|
  x.report("select.any?") do
    arr.select {|x| x > 1 }.any?
  end

  x.report("any?") do
    arr.any? {|x| x > 1 }
  end

  x.compare!
end
```

```
Comparison:
                any?:  4845559.5 i/s
         select.any?:  2671012.3 i/s - 1.81x  slower
```

Similar results for other 2 methods.

Tested on `rubocop`, `rails` and `gitlab` - found 2 offenses on `gitlab`.

Another possible cop name I can think of - `SelectFollowedByPredicate`.


vlad-pisanov:
Great idea for a cop! 

Should `.one?` be included as well? `[1, 2, 3].one? { |x| x > 2 }`

Perhaps the list of predicates could be configurable? I realize that `[1, 2, 3].empty? { |x| x > 3 }` is more performant, but it just doesn't look natural to me. 😬

fatkodima:
@koic Wdyt on this one? If ok, I will update it with the suggestion.

koic:
> So do you think a configurable list, as suggested, wouldn't be better?

I'm not sure if there are other unknown methods for configure. And creating it configurable means providing an API (the promised name for user). e.g. `PredicateMethods`.
So, It might be better to start with `ActiveSupportExtensionsEnabled` and then provide configurability when needed. This approach does not provide extension points for the user, but it also doesn't require providing an API name.

So, I feel like `ActiveSupportExtensionsEnabled` is enough this time.

PR #371, opened 2023-09-14
[Fix #370] Fix an incorrect autocorrect for `Performance/RedundantMatch`

### LINKED ISSUE #370: Incorrect autocorrect for `Performance/RedundantMatch` cop
I've noticed that the cop's autocorrection is incorrect when using operators with lower precedence than `=~` as an argument to `match`.

Here's the reproducible code:

```ruby
arg = nil
puts "pass" if /regex/.match(arg || "default")
```

--------

## Expected behavior

To preserve the original behavior, parentheses should be added around the RHS.

```ruby
arg = nil
puts "pass" if /regex/ =~ (arg || "default")
```

This code produces no output.

## Actual behavior

Parentheses are missing, which results in the condition being evaluated as `(/regex/ =~ arg) || "default"`.

```ruby
arg = nil
puts "pass" if /regex/ =~ arg || "default"
```

This code outputs "pass".

1. Save the provided code snippet into a file named `example.rb`.
2. Execute `bundle exec rubocop --only Performance/RedundantMatch -a example.rb`.

## RuboCop version

```
$ bundle exec rubocop -V
1.56.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.0.2) [
x86_64-linux]
  - rubocop-performance 1.19.0
  - rubocop-rspec 2.22.0
```

ymap:
Fixes #370.

This PR fixes an incorrect autocorrect for `Performance/RedundantMatch` when expressions with lower precedence than `=~` are used as an argument.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/performance/redundant_match.rb
+++ b/lib/rubocop/cop/performance/redundant_match.rb
@@ -23,6 +23,8 @@ module RuboCop
         MSG = 'Use `=~` in places where the `MatchData` returned by `#match` will not be used.'
         RESTRICT_ON_SEND = %i[match].freeze
 
+        HIGHER_PRECEDENCE_OPERATOR_METHODS = %i[| ^ & + - * / % ** > >= < <= << >>].freeze
+
         # 'match' is a fairly generic name, so we don't flag it unless we see
         # a string or regexp literal on one side or the other
         def_node_matcher :match_call?, <<~PATTERN
@@ -47,7 +49,7 @@ module RuboCop
         private
 
         def autocorrect(corrector, node)
-          new_source = "#{node.receiver.source} =~ #{node.first_argument.source}"
+          new_source = "#{node.receiver.source} =~ #{replacement(node)}"
 
           corrector.replace(node, new_source)
         end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/performance/redundant_match.rb
```diff
@@ -57,6 +59,35 @@ def autocorrectable?(node)
...
+            arg.source
+          end
+        end
+
+        def requires_parentheses?(arg)
+          return true if arg.if_type? && arg.ternary?
+          return true if arg.and_type? || arg.or_type? || arg.range_type?
+          return requires_parentheses_for_call_like?(arg) if call_like?(arg)
+
+          false
```
koic, line 78: Can you refactor?

```suggestion
          call_like?(arg) && requires_parentheses_for_call_like?(arg)
```

lib/rubocop/cop/performance/redundant_match.rb
```diff
@@ -57,6 +59,35 @@ def autocorrectable?(node)
...
+        end
+
+        def requires_parentheses_for_call_like?(arg)
+          return false if arg.parenthesized? || !arg.arguments?
+          return false if HIGHER_PRECEDENCE_OPERATOR_METHODS.include?(arg.method_name)
+
+          true
```
koic, line 85:
```suggestion
          return false if arg.parenthesized? || !arg.arguments?

          !HIGHER_PRECEDENCE_OPERATOR_METHODS.include?(arg.method_name)
```

ymap:
@koic Thanks for the suggestions! Amended the changes.

koic:
Thanks!

PR #378, opened 2023-10-22
Update the doc for `Performance/Count`

ryym:
This pull request improves the doc for `Performance/Count` .

- Removed almost duplicate sentences (left the latter one only):
    > ActiveRecord's `count` ignores the block that is passed to it.

    > `ActiveRecord` will ignore the block that is passed to `count`.

- Added Rails version that fixed the described `count` method issue. Since it has been fixed in 2016, newer Rails users will not encounter this issue. 
    > Fixed ActiveRecord::Relation#count to use Ruby's Enumerable#count for counting records when a block is passed as argument instead of silently ignoring the passed block. ([Pull Request](https://github.com/rails/rails/pull/24203))
    - <https://guides.rubyonrails.org/5_1_release_notes.html>

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/performance/count.rb
+++ b/lib/rubocop/cop/performance/count.rb
@@ -9,8 +9,7 @@ module RuboCop
       #
       # @safety
       #   This cop is unsafe because it has known compatibility issues with `ActiveRecord` and other
-      #   frameworks. ActiveRecord's `count` ignores the block that is passed to it.
-      #   `ActiveRecord` will ignore the block that is passed to `count`.
+      #   frameworks. Before Rails 5.1, `ActiveRecord` will ignore the block that is passed to `count`.
       #   Other methods, such as `select`, will convert the association to an
       #   array and then run the block on the array. A simple work around to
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/cops_performance.adoc
```diff
@@ -544,8 +544,7 @@ passed to the `count` call.
 === Safety
 
 This cop is unsafe because it has known compatibility issues with `ActiveRecord` and other
-frameworks. ActiveRecord's `count` ignores the block that is passed to it.
-`ActiveRecord` will ignore the block that is passed to `count`.
+frameworks. Before Rails 5.1, `ActiveRecord` will ignore the block that is passed to `count`.
```
koic, line 547: This adoc file is auto-generated, so no changes are necessary.

koic:
Can you squash your commits into one?

ryym:
Thank you for your reviewing!
I squashed the commits and discarded the change of adoc file.

PR #381, opened 2023-11-11
Revise links

ydakuka:
1) Follow by https://github.com/fastruby/fast-ruby/pull/222
2) I've replaced invalid links with valid ones.


koic:
Good catch!

PR #392, opened 2023-11-25
Fix a false negative for `Performance/ReverseEach` when safe navigation is between `reverse` and `each`

fatkodima:
Fixes https://github.com/rubocop/rubocop-performance/issues/390.

### DIFF
--- a/lib/rubocop/cop/performance/reverse_each.rb
+++ b/lib/rubocop/cop/performance/reverse_each.rb
@@ -27,7 +27,7 @@ module RuboCop
         RESTRICT_ON_SEND = %i[each].freeze
 
         def_node_matcher :reverse_each?, <<~MATCHER
-          (send (call _ :reverse) :each)
+          (call (call _ :reverse) :each)
         MATCHER
 
         def on_send(node)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/performance/reverse_each.rb
```diff
@@ -27,7 +27,7 @@ class ReverseEach < Base
         RESTRICT_ON_SEND = %i[each].freeze
 
         def_node_matcher :reverse_each?, <<~MATCHER
-          (send (call _ :reverse) :each)
+          ({send csend} (call _ :reverse) :each)
```
koic, line 30:
```suggestion
          (call (call _ :reverse) :each)
```

koic:
CI has been failed. Can you rebase with the latest master branch?

fatkodima:
Updated.

spec/rubocop/cop/performance/reverse_each_spec.rb
```diff
@@ -12,6 +12,8 @@
     expect_offense(<<~RUBY)
       array&.reverse.each { |e| puts e }
              ^^^^^^^^^^^^ Use `reverse_each` instead of `reverse.each`.
+      array&.reverse&.each { |e| puts e }
+             ^^^^^^^^^^^^^ Use `reverse_each` instead of `reverse.each`.
```
koic, line 16: Please separate this, an independent test instead of mixing it in the existing `it` block.

fatkodima: Hate inventing new names for tests 😅 Updated

PR #423, opened 2023-12-12
[Fix #421] Fix false positives for `Performance/TimesMap`

### LINKED ISSUE #421: Fix bug in the `Performance/TimesMap` cop
Follow by https://github.com/rubocop/rubocop-performance/pull/420/#discussion_r1421360806

### LINKED ISSUE #422: Improve the `Performance/TimesMap` cop

```ruby
# bad
(0..8).map { |i| i }
0.upto(8).map { |i| i }

# good
Array.new(9) { |i| i }
```

koic:
Fixes #421.

This PR fixes false positives for `Performance/TimesMap` with a block with safe navigation call for nil literal.

It does not have a changelog entry because it is a patch to a feature that has not been released yet.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/performance/times_map_spec.rb
+++ b/spec/rubocop/cop/performance/times_map_spec.rb
@@ -29,6 +29,48 @@ RSpec.describe RuboCop::Cop::Performance::TimesMap, :config do
         end
       end
 
+      context 'with a block with safe navigation call for integer literal' do
+        it 'registers an offense and corrects' do
+          expect_offense(<<~RUBY, method: method)
+            42&.times&.#{method} { |i| i.to_s }
+            ^^^^^^^^^^^^{method}^^^^^^^^^^^^^^^ Use `Array.new(42)` with a block instead of `.times.#{method}`.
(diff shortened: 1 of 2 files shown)


spec/rubocop/cop/performance/times_map_spec.rb
```diff
@@ -29,6 +29,32 @@
...
+      end
+
+      context 'with a block with safe navigation call for nil literal' do
+        it 'does not register an offense and corrects' do
+          expect_no_offenses(<<~RUBY, method: method)
+            nil&.times&.#{method} { |i| i.to_s }
+          RUBY
+        end
+      end
+
```
ydakuka, line 57: What about this case?

```ruby
it 'does not register an offense and corrects' do
  expect_no_offenses(<<~RUBY, method: method)
    foo&.times&.#{method} { |i| i.to_s }
  RUBY
end
```

koic: Sure, I've fixed that case.

PR #427, opened 2023-12-21
Remove `const_get` from `StringIdentifierArgument` target

### LINKED ISSUE #425: The autocorrection for Performance/StringIdentifierArgument breaks code
## Expected behavior

The cop's autocorrection should not break code.

## Actual behavior

The cop's autocorrection breaks code.

The following code runs:

```
# frozen_string_literal: true

module MyNamespace
  class MyClass
  end
end

prefix = 'MyNamespace::'
class_name = 'MyClass'
Kernel.const_get("#{prefix}#{class_name}")
```

Running rubocop on it transforms it to this:

```
# frozen_string_literal: true

module MyNamespace
  class MyClass
  end
end

prefix = 'MyNamespace::'
class_name = 'MyClass'
Kernel.const_get(:"#{prefix}#{class_name}")
```

Running the corrected code produces the following error:
```
app.rb:10:in `const_get': wrong constant name MyNamespace::MyClass (NameError)

Kernel.const_get(:"#{prefix}#{class_name}")
      ^^^^^^^^^^
	from app.rb:10:in `<main>'
```

## RuboCop version
```
1.59.0 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.1.1) [x86_64-linux]
  - rubocop-performance 1.20.0
```

okuramasafumi:
[Fix #425]
`const_get` doesn't accept symbols that are something like `:"Foo::Bar"`, and that's intended according to this issue. https://bugs.ruby-lang.org/issues/12319
Therefore, changing String to Symbol might not work as expected. As far as I know that's related to `const_get` only so we can remove it from the target of `StringIdentifierArgument`

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_string_identifier_argument_with_const_get.md
+++ b/changelog/fix_string_identifier_argument_with_const_get.md
@@ -1,0 +1,2 @@
+* [#427](https://github.com/rubocop/rubocop-performance/issues/425): Remove `const_get` from `StringIdentifierArgument` target. ([@okuramasafumi][])
+
--- a/lib/rubocop/cop/performance/string_identifier_argument.rb
+++ b/lib/rubocop/cop/performance/string_identifier_argument.rb
@@ -38,7 +38,7 @@ module RuboCop
         # And `attr` may not be used because `Style/Attr` registers an offense.
         # https://github.com/rubocop/rubocop-performance/issues/278
         RESTRICT_ON_SEND = (%i[
-          class_variable_defined? const_defined? const_get const_set const_source_location
+          class_variable_defined? const_defined? const_set const_source_location
           define_method instance_method method_defined? private_class_method? private_method_defined?
           protected_method_defined? public_class_method public_instance_method public_method_defined?
           remove_class_variable remove_method undef_method class_variable_get class_variable_set


Earlopain:
I don't think this is the right solution.

This should just skip interpolated strings for `const_get`. In addition it's easily analyzable if the string literal contains double-colons to prevent a false positive.

From testing, this also applies to:
* `const_source_location`
* `const_defined?`

These don't:
* `public_const`/`private_const` don't accept nested constants
* `deprecate_constant` as well
* Same for `remove_const` and `set_const`

You also need to update tests.

okuramasafumi:
@Earlopain You are right. At least this change must cover all affected methods.
One concern is that the rule will be a little complex. Some might be confused about why it's not applied for some cases. We need to have a good documentation to explain.

okuramasafumi:
@Earlopain What about this case:

```ruby
const_get("#{process(name)}Resource")
```

Here, `process` method returns a string that MIGHT contain `::`. Is it analyzable?

Earlopain:
Perhaps in some trivial cases but generally no. I would just ignore those.

koic:
I'll close this PR in favor of #430, which proposals a better resolution. Thank you.

PR #433, opened 2024-01-12
Add new `ArrayPushSingle` cop to replace `a.push(x)` with `a << x`

### LINKED ISSUE #431: Cop idea: prefer `<<` over `push`
`a << i` is faster than `a.push(e)`. I suspect it's because `#push` takes `*args`, and the overhead from that is relatively slower than `<<`, which takes a single arg.

```
                push     2.560M i/100ms
                  <<     3.169M i/100ms
                push     25.715M (± 1.7%) i/s -    130.557M in   5.078559s
                  <<     31.909M (± 1.0%) i/s -    161.598M in   5.064825s
```

<details><summary><code>benchmark.rb</code></summary>

```ruby
a = []
    
Benchmark.ips do |x|
  x.report("push") do |times|
    i = 0
    a.clear
    while (i += 1) < times
      a.push(i)
    end
  end
  
  x.report("<<") do |times|
    i = 0
    a.clear
    while (i += 1) < times
      a << i
    end
  end
  
  x.compare!
end
```

</details>

amomchilov:
Closes #431

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -11,6 +11,13 @@ Performance/AncestorsInclude:
   Safe: false
   VersionAdded: '1.7'
 
+Performance/ArrayPushSingle:
+  Description: 'Use `array << x` instead of `array.push(x)`.'
+  Reference: 'https://github.com/fastruby/fast-ruby#ancestorsinclude-vs--code'
+  Enabled: true
```
amomchilov, line 17: Should it be `true`, `false` or `pending`?

lib/rubocop/cop/performance/redundant_split_regexp_argument.rb
```diff
@@ -48,7 +48,7 @@ def replacement(regexp_node)
           stack = []
           chars = regexp_content.chars.each_with_object([]) do |char, strings|
             if stack.empty? && char == '\\'
-              stack.push(char)
+              stack.push(char) # rubocop:disable Performance/ArrayPushSingle
```
amomchilov, line 51: Keeping `push` is nicer here, to match `pop` below

spec/rubocop/cop/performance/array_push_single_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    RUBY
+  end
+
+  it 'registers an offense and corrects when using `push` with a single element and safe navigation operator' do
+    expect_offense(<<~RUBY)
+      array&.push(element)
+      ^^^^^^^^^^^^^^^^^^^^ Use `<<` instead of `push`.
+    RUBY
+
+    # gross. TODO: make a configuration option?
```
amomchilov, line 21: What's the best way to handle this?

lib/rubocop/cop/performance/array_push_single.rb
```diff
@@ -0,0 +1,56 @@
...
+        def_node_matcher :push_call?, <<~PATTERN
+          (call $_receiver $%PUSH_METHODS $!(splat _))
+        PATTERN
+
+        def on_send(node)
+          push_call?(node) do |receiver, method_name, element|
+            message = format(MSG, current: method_name)
+
+            add_offense(node, message: message) do |corrector|
+              corrector.replace(node, "#{receiver.source} << #{element.source}")
```
vlad-pisanov, line 39: there are some edge cases that this won't handle correctly e.g.:
```ruby
# original
[].push(1 && 2)  # result is [2]
# corrected
[] << 1 && 2     # result is 2

# original
[].push 1 << 2   # result is [4]
# corrected
[] << 1 << 2     # result is [1,2]

# original
[].push(1) + [].push(2)  # result is [1,2]
# corrected
[] << 1 + [] << 2
TypeError: Array can't be coerced into Integer
```

spec/rubocop/cop/performance/array_push_single_spec.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Performance::ArrayPushSingle, :config do
+  it 'registers an offense and corrects when using `push` with a single element' do
+    expect_offense(<<~RUBY)
+      array.push(element)
+      ^^^^^^^^^^^^^^^^^^^ Use `<<` instead of `push`.
```
vlad-pisanov, line 7: should probably add a test for `append` as well

spec/rubocop/cop/performance/array_push_single_spec.rb
vlad-pisanov: I'd just ignore it. `&.<<` looks horrible

amomchilov: Agree.

lib/rubocop/cop/performance/array_push_single.rb
amomchilov: Hmm does Rubocop have any utilities for detecting the precedence of a subexpr, relative to the other parts of the expression it's in?

lib/rubocop/cop/performance/array_push_single.rb
```diff
@@ -0,0 +1,56 @@
...
+      #
+      # @example
+      #   # bad
+      #   array.push(1)
+      #
+      #   # good
+      #   array << 1
+      #
+      #   # good
+      #   array.push(1, 2, 3) # `<<` only works for one element
```
ydakuka, line 20: > `<<` only works for one element

```ruby
2.7.6 :001 > arr = []
 => [] 
2.7.6 :002 > arr << 1 << 2 << 3
 => [1, 2, 3] 
```

ydakuka:
Is this related to https://github.com/rubocop/rubocop/issues/12053 ?

amomchilov:
@ydakuka I hadn't seen that, but I guess it's pretty similar!

I think two separate cops would be good here. One to turn `a.push(x)` into `a << x`, and another to take a series of `a << x`s (or pushes) and turn them into one.

config/default.yml
bdewater: The new cop generator should output pending. On the next major release all pending cops are enabled in bulk according to the [docs](https://docs.rubocop.org/rubocop/versioning.html#pending-cops).

PR #439, opened 2024-01-31
[Fix #437] Fix a false positive for `Performance/ChainArrayAllocation`

### LINKED ISSUE #437: Incorrect recommendation suggested by Performance/ChainArrayAllocation Cop
## Expected behavior

RuboCop suggests valid corrections on arrays.

## Actual behavior

Identifies multiple uses of ActiveRecord query `select` [method](https://apidock.com/rails/ActiveRecord/QueryMethods/select)
While calling `select` multiple times on the same relation may technically be redundant, this is a completely distinct method from `Array.select`, and thus suggesting to use `select!` (which does not exist), should not happen.

Run rubocop on this file with `Performance/ChainArrayAllocation` enabled

models/test.rb
```
class Test < ActiveRecord::Base

  scope :test_scope, ->  {
    select(:field_one, :field_two).
    select("COUNT(DISTINCT(tests.field_three)) AS num").
    where("tests.field_one IS NOT NULL").
    group(:field_two)
  }

  def self.something
    Test.all
        .select(:field_one, :field_two)
        .select("COUNT(DISTINCT(tests.field_three)) AS num")
        .where("tests.field_one IS NOT NULL")
        .group(:field_two)
  end

end
```

## RuboCop version

Found on version
```
1.31.0 (using Parser 3.3.0.5, rubocop-ast 1.30.0, running on ruby 2.7.8 x86_64-darwin21)
  - rubocop-performance 1.19.1
  - rubocop-rails 2.15.2
```

Still occurs on latest
```
1.60.2 (using Parser 3.3.0.5, rubocop-ast 1.30.0, running on ruby 2.7.8) [x86_64-darwin21]
  - rubocop-performance 1.20.2
  - rubocop-rails 2.23.1
```

----------

Perhaps this is why this cop is disabled by default, because it has these are unsolvable false positives?

koic:
Fixes #437.

This PR fixes a false positive for `Performance/ChainArrayAllocation` when using `select` with block argument after `select`.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/performance/chain_array_allocation_spec.rb
+++ b/spec/rubocop/cop/performance/chain_array_allocation_spec.rb
@@ -53,4 +53,38 @@ RSpec.describe RuboCop::Cop::Performance::ChainArrayAllocation, :config do
       RUBY
     end
   end
+
+  describe 'when using `select` with block argument after `select` with positional arguments' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        model.select(:foo, :bar).select { |item| item.do_something }
+      RUBY
+    end
+  end
+
+  describe 'when using `select` with block argument after `select` with block argument' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        model.select { |item| item.foo }.select { |item| item.bar }
+                                        ^^^^^^^ Use unchained `select` and `select!` (followed by `return array` if required) instead of chaining `select...select`.
+      RUBY
+    end
+  end
+
+  describe 'when using `select` with block argument after `select` with numbered block argument' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        model.select { _1.foo }.select { |item| item.bar }
+                               ^^^^^^^ Use unchained `select` and `select!` (followed by `return array` if required) instead of chaining `select...select`.
+      RUBY
+    end
+  end
+
+  describe 'when using `select` with positional arguments after `select`' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        model.select(:foo, :bar).select(:baz, :qux)
+      RUBY
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/performance/chain_array_allocation_spec.rb
```diff
@@ -53,4 +53,21 @@
...
   end
+
+  describe 'when using `select` with block argument after `select`' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        model.select(:foo, :bar).select { |item| item.do_something }
+                                ^^^^^^^ Use unchained `select` and `select!` (followed by `return array` if required) instead of chaining `select...select`.
+      RUBY
+    end
+  end
```
matthewhively, line 64: Forgive my possible stupidity (I'm not terribly familiar with how the tests in rubocop work)
But wouldn't this still be wrong?
Wouldn't it flag `select` (QueryMethods#select) followed by a `select` (enumerable).
Which isn't a double select that requires optimization?

You would however want to flag the 3rd select here right?
`model.select(:foo, :bar).select { |item| item.do_something }.select { |item| item.do_something_else }`

koic: I'm not entirely sure, this patch works as follows:

```ruby
it 'registers an offense' do
  expect_offense(<<~RUBY)
    model.select(:foo, :bar).select { |item| item.do_something }.select { |item| item.do_something_else }
                            ^^^^^^^ Use unchained `select` and `select!` (followed by `return array` if required) instead of chaining `select...select`.
                                                                ^^^^^^^ Use unchained `select` and `select!` (followed by `return array` if required) instead of chaining `select...select`.
  RUBY
end
```

So, this is a patch that distinguishes between `select(:arg)` and `select { |x| x }`. What potential issues might arise from this?

matthewhively: So, let me rephrase
with this code 
```
expect_offense(<<~RUBY)
        model.select(:foo, :bar).select { |item| item.do_something }
                                ^^^^^^^ Use unchained `select` and `select!` (followed by `return array` if required) instead of chaining `select...select`.
RUBY
```
What should the suggested corrected code look like?

koic: Ah, I understand. This should definitely be recognized as a false positive and I've fixed it. Thank you!

PR #440, opened 2024-02-06
Add Performance::NumericPredicate cop

miry:
Performance::NumericPredicate cop identifies places where numeric uses predicates like `positive?`, `negative?` and for some cases `zero?` should be converted to compare operator.

The `Performance::NumericPredicate` cop is added to identify instances where numeric predicates such as `positive?`, `negative?`, and occasionally `zero?` should be replaced with comparison operators for improved efficiency.

Predicates incur a performance overhead by executing a method before comparison. A small benchmark comparison between using a comparison operator (`> 0`) and `positive?` illustrates the performance difference:

```ruby
x.report("compare with 0") { arr.each {|i| i > 0 } }
x.report("positive?") { arr.each {|i| i.positive? } }
```

Benchmark results on Ruby 3.3.0 (with YJIT) indicate a significant performance gain when using the comparison operator:

```
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin23]
      compare with 0     1.000 i/100ms
           positive?     1.000 i/100ms
      compare with 0      3.153 (± 0.0%) i/s -     95.000 in  30.132600s
           positive?      2.397 (± 0.0%) i/s -     72.000 in  30.042688s

Comparison:
      compare with 0:        3.2 i/s
           positive?:        2.4 i/s - 1.32x  slower
```

This cop is unsafe because it cannot be guaranteed that the receiver is Number and could be noisy.

**Replace this text with a summary of the changes in your PR.
The more detailed you are, the better.**

-----------------

* [ ] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [ ] Wrote [good commit messages][1].
* [ ] Added tests.
* [ ] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.


lib/rubocop/cop/performance/numeric_predicate.rb
```diff
@@ -0,0 +1,61 @@
...
+      #
+      class NumericPredicate < Base
+        extend AutoCorrector
+
+        MSG = 'Use compare operator `%<good>s` instead of `%<bad>s`.'
+        RESTRICT_ON_SEND = %i[positive? zero? negative?].freeze
+        REPLACEMENTS = { negative?: '<', positive?: '>', zero?: '==' }.freeze
+
+        def_node_matcher :num_predicate?, <<~PATTERN
+          (send $numeric_type? ${:negative? :positive? :zero?})
```
bdewater, line 33: Should this also look for [`nonzero?`](https://docs.ruby-lang.org/en/3.0/Numeric.html#method-i-nonzero-3F)

Earlopain:
This is just [`Style/NumericPredicate`](https://docs.rubocop.org/rubocop/cops_style.html#stylenumericpredicate), is it not? It's configured for predicates by default so it would say that they will flip-flop between each other if not handled specifically. I'm also not sure if its worth the effort to duplicate a cop for what should basically be a different config value.

miry:
@Earlopain Are there any established practices for handling conflicts between rubocop cops and performance cops?
I wonder, if there is a way to check the RuboCop configuration to ensure that the performance-related cops are enabled.

PR #457, opened 2024-08-21
Don't use deprecated `Cop.registry` in specs

Earlopain:
Also update cop filtering for https://github.com/rubocop/rubocop/pull/13123


koic:
Thanks!

PR #398, opened 2023-11-27
Migrate to GitHub Actions from CircleCI

koic:
This PR will centralize CI to GitHub Actions as introduced in https://github.com/rubocop/rubocop-performance/pull/388.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop-performance/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #458, opened 2024-08-21
Fix a random test failure for `Performance/Sum`

Earlopain:
The skipped tests fail with the following command: `bundle exec rspec ./spec/rubocop/cli/autocorrect_spec.rb[1:1] ./spec/rubocop/cop/performance/sum_spec.rb[1:13] --seed 26805` (or maybe not, there is something extra going on I don't understand)

This is fine to do in the rubocop main repository but here we inject additional config values that are not restored.

There are some things going on that I don't entirely understand:
* Why only failures in this one file
* Why did koic not manage to reproduce with the exact same seed
* Why did the failures seemingly disappear and reappear without any changes to the code

But regardless, this is more correct. Removing the JRuby skips is a bit of a guess on my side but I don't see what else should prevent them from passing. Let's see how it goes.

Bit of extra context starting at https://github.com/rubocop/rubocop-performance/pull/398#issuecomment-2301743107

### DIFF
--- a/spec/rubocop/cli/autocorrect_spec.rb
+++ b/spec/rubocop/cli/autocorrect_spec.rb
@@ -7,10 +7,16 @@ RSpec.describe 'RuboCop::CLI --autocorrect', :isolated_environment do # rubocop:
 
   include_context 'mock console output'
 
-  before do
+  # Restore injected config as well
+  around do |ex|
+    previous_config = RuboCop::ConfigLoader.default_configuration
+
     RuboCop::ConfigLoader.default_configuration = nil
     RuboCop::ConfigLoader.default_configuration.for_all_cops['SuggestExtensions'] = false
     RuboCop::ConfigLoader.default_configuration.for_all_cops['NewCops'] = 'disable'
+    ex.call
+  ensure
+    RuboCop::ConfigLoader.default_configuration = previous_config
   end
 
   it 'corrects `Performance/ConstantRegexp` with `Performance/RegexpMatch`' do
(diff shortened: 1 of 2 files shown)


koic:
> Why did koic not manage to reproduce with the exact same seed

When I ran it again, the following error occurred:

```console
$ bundle exec rspec './spec/rubocop/cli/autocorrect_spec.rb[1:1]' './spec/rubocop/cop/performance/sum_spec.rb[1:13]' --seed 26805
Run options: include {:ids=>{"./spec/rubocop/cli/autocorrect_spec.rb"=>["1:1"], "./spec/rubocop/cop/performance/sum_spec.rb"=>["1:13"]}}

Randomized with seed 26805
.F

Failures:

  1) RuboCop::Cop::Performance::Sum autocorrects `&:+` when initial value is not provided
     Failure/Error:
       expect_correction(<<~RUBY)
         array.sum
       RUBY

     RuntimeError:
       Expected correction but no corrections were made
     # ./spec/rubocop/cop/performance/sum_spec.rb:278:in 'block (3 levels) in <top (required)>'

Finished in 0.39425 seconds (files took 2.33 seconds to load)
2 examples, 1 failure

Failed examples:

rspec './spec/rubocop/cop/performance/sum_spec.rb[1:13]' # RuboCop::Cop::Performance::Sum autocorrects `&:+` when initial value is not provided

Randomized with seed 26805
```

Earlopain:
Ah, right. Yeah, that is what I was getting as well. Still, I would expect this to come up all the for other cops as well and I don't know why `Sum` is special here.

PR #462, opened 2024-09-05
Add new `ZipWithoutBlock` cop

corsonknowles:
Performance Cop for the more efficient way to generate an Array of Arrays.

* Performs 40-90% faster than`.map` to iteratively wrap array contents. 
* Performs 5 - 55% faster on converting a range to an array of arrays with `.map`, depending on size.
* Performs faster than mutating an array into an array of arrays in place with `.map!`
* Performs faster than using `each_with_object`

This optimization is particularly helpful in performance sensitive paths or in large scale applications that need to generate arrays of arrays at scale. For example, leveraging the bulk enqueuing feature in Sidekiq requires an array of arrays, and is by definition used at scale.

A performance gain is always present for all sizes of arrays and ranges. The gain is smallest with small ranges, but still significant for small arrays.

This is not a style cop, but it is poetic that the more performant approach also has simpler and shorter syntax. 

`.zip` has been intentionally optimized in Ruby. This has been discussed publicly since at least 2012:
* https://stackoverflow.com/a/5254764

Official `.zip` documentation: 
* https://ruby-doc.org/3.3.5/Array.html#method-i-zip
* https://ruby-doc.org/3.3.5/Enumerable.html#method-i-zip

Source code for `.zip`:
* https://github.com/ruby/ruby/blob/v3_3_0/array.c#L4418
* https://github.com/ruby/ruby/blob/v3_3_0/enum.c#L3448

This performance cop isn't just an announcement to use `.zip` in the spirit of appreciating Ruby's great features, it is also a useful and necessary tool to leverage Rubocop to clean up and add rigor in large Ruby code bases. Rubocop is a much better approach than pattern matching for clean up at scale here, and it comes with the added benefit of proactive user feedback as additions are made to the code base going forward.

For Arrays with 1000 entries, a common size for bulk operations, this performs 70% faster. Here it is at 70% faster, using `benchmark-ips`:

```shell
gem install benchmark-ips
```
```ruby
require 'benchmark'
require 'benchmark/ips'

array = (1..1000).to_a; Benchmark.ips do |x|
  x.report(".zip:") do
    array.zip
  end
  
  x.report(".map { |id| [id] }:") do
    array.map { |id| [id] }
  end
end
```

```bash
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin23]
               .zip:     4.019k i/100ms
 .map { |id| [id] }:     2.404k i/100ms
               .zip:     41.193k (± 1.4%) i/s -    208.988k in   5.074349s
 .map { |id| [id] }:     24.277k (± 2.8%) i/s -    122.604k in   5.054623s
```

-----------------

* [x] Added tests.


corsonknowles:
Some of the gain here comes from simply not allocating the block. But I think most of it comes from the lower level conversion of the array into an array of arrays.

corsonknowles:
Dear @koic and @Earlopain, 
Are there any questions I can answer for you or modifications I can provide? 

Thank you so much for your consideration!

changelog/new_merge_pull_request_459_from.md
```diff
@@ -0,0 +1 @@
+* [#462](https://github.com/rubocop/rubocop-performance/pull/462): Merge pull request #462 from corsonknowles:add_performance_use_zip_to_wrap_arrays. ([@corsonknowles][])
```
Earlopain, line 1: Please write a meaningful message

config/default.yml
```diff
@@ -372,3 +372,9 @@ Performance/UriDefaultParser:
   Description: 'Use `URI::DEFAULT_PARSER` instead of `URI::Parser.new`.'
   Enabled: true
   VersionAdded: '0.50'
+
+Performance/UseZipToWrapArrayContents:
+  Description: 'Checks for `.map { |id| [id] }` and suggests replacing it with `.zip`.'
+  Enabled: true
```
Earlopain, line 378:
```suggestion
  Enabled: pending
```

Cops are not enabled in minor versions

config/default.yml
```diff
@@ -372,3 +372,9 @@ Performance/UriDefaultParser:
...
+  SafeAutoCorrect: true
```
Earlopain, line 379:
```suggestion
```

This should already be the default

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,87 @@
...
+        # @!method map_with_array_numblock?(node)
+        def_node_matcher :map_with_array_numblock?, <<~PATTERN
+          (numblock
+            (send _ :map)
+            1
+            (array (lvar _))
+          )
+        PATTERN
+
+        def on_block(node)
```
Earlopain, line 53: Can you use `on_send` with `RESTRICT_ON_SEND = [:zip]` instead? That will have better performance.

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
```diff
@@ -0,0 +1,170 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Performance::UseZipToWrapArrayContents, :config do
+  context 'when using map with array literal' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        [1, 2, 3].map { |id| [id] }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `.zip` instead of `.map { |id| [id] }`.
```
Earlopain, line 8: This offense range is a bit too long. It should only start at `.map`.

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+      RUBY
+    end
+
+    it 'corrects the code' do
+      new_source = autocorrect_source('[1, 2, 3].map { |id| [id] }')
```
Earlopain, line 13: Use `expect_correction` and fold it into the `expect_offense` test. Same for all the others.

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+  context 'when using map with a short iterator name' do
+    let(:source) do
+      <<~RUBY
+        [1, 2, 3].map { |e| [e] }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^ Use `.zip` instead of `.map { |id| [id] }`.
+      RUBY
+    end
+
+    it 'registers an offense' do
+      expect_offense(source)
```
Earlopain, line 28: Use `expect_offense(<<~RUBY)` instead of putting it in `let`. Thats largely the convention  over the codebases. Same for other instances.

Earlopain (commented):
This largely makes sense to me. Never really thought about using `zip` without arguments, that's a nice idea.

I think this pattern is relatively common in rails where they need to handle composite primary keys. I'll edit my findings in later if I decide to check.

Tip: use `benchmark-ips` for future benchmarks. The output is much more legible.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
corsonknowles: I got this working to replace on_block so its only on map blocks with  
RESTRICT_ON_SEND = [:map]

But I am stymied on how to handle `on_numblock` this way. Ideas? I can try to stitch the two different approaches together.

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Thanks! What's the mechanism for shortening the offense range?

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: I got this pattern from here: https://github.com/search?q=repo%3Arubocop%2Frubocop-performance%20new_source&type=code 
It looks common in the recent commits. Did I misapply it?

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Thanks, will do!

changelog/new_merge_pull_request_459_from.md
corsonknowles: Done, will push it all up as 1 over-write.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
corsonknowles: Actually the distinct handling works quite nicely:

```
        def on_send(node)
          return unless node.method?(:map)

          parent = node.parent
          return unless parent

          register_offense(parent) if parent.block_type? && map_with_array?(parent)
        end

        def on_numblock(node)
          return unless node.method?(:map)
          return unless map_with_array_numblock?(node)

          register_offense(node)
        end
```

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: And fixed!

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Okay, fixed! I see it now, the other standalone tests aren't already matching the same string to expect an offense

corsonknowles:
I really appreciate the thoughtful comments @Earlopain 

Fully revised and ready for review. 

I had run the original version on 50,000 files. I added 2 more tests and diversified the input slightly to get confident in the new on_send approach. I predict this should be safe to merge now.

corsonknowles:
Thanks for the `benchmark-ips` tip. I added some data using it for large and small arrays.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,87 @@
...
+          register_offense(parent) if parent.block_type? && map_with_array?(parent)
+        end
+
+        def on_numblock(node)
+          return unless node.method?(:map)
+          return unless map_with_array_numblock?(node)
+
+          register_offense(node)
+        end
+
```
Earlopain, line 69:
```suggestion
          register_offense(parent) if map_with_array?(parent) || map_with_array_numblock?(parent)
```

I'm sure you can combine the pattern matchers somehow but I'm not good enough with them to tell how

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,87 @@
...
+        def_node_matcher :map_with_array_numblock?, <<~PATTERN
+          (numblock
+            (send _ :map)
+            1
+            (array (lvar _))
+          )
+        PATTERN
+
+        def on_send(node)
+          return unless node.method?(:map)
```
Earlopain, line 55:
```suggestion
```

This is not needed, `RESTRICT_ON_SEND` ensures it already

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Performance::UseZipToWrapArrayContents, :config do
+  context 'when using map with array literal' do
+    it 'registers an offense and corrects to use zip with no arguments' do
+      expect_offense(<<~RUBY)
+        [1, 2, 3].map { |id| [id] }
+                  ^^^ Use `.zip` instead of `.map { |id| [id] }`.
```
Earlopain, line 8:
```suggestion
                  ^^^^^^^^^^^^^^^^^ Use `.zip` instead of `.map { |id| [id] }`.
```

I think this should include the block as well

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,87 @@
...
+      # @example
+      #   # bad
+      #   [1, 2, 3].map { |id| [id] }
+      #
+      #   # good
+      #   [1, 2, 3].zip
+      #
+      # @example
+      #   # bad
+      #   [1, 2, 3].map { [_1] }
```
Earlopain, line 17: I don't believe this needs an explicit callout about numblocks in the example here. It should already be implied

Earlopain (commented):
Almost done from my side, just a few remaining things.

Can you also add a test for `map` and `map {}`? These are common places for errors in the past. I think it works fine in your case right now

technicalpickles:
First off, this will be a nice improvement for our code base, as we often use and recommend the `map {|id| [id]}` for `Sidekiq.perform_bulk`. Disclaimer: work with @corsonknowles 😁 

I ran some of my own benchmarks based on @corsonknowles's in the body against small (4), medium (1000) and big (50k) arrays using benchmark-ips, and also added benchmark-memory for good measure. High level, we have:

- small: zip is 1.45x faster
- medium: zip is 1.74x faster
- big: zip is 1.93x faster

I'm kinda surprised, but memory usage ends up the same which I'm not quite sure how to explain yet 😅

Here is [the benchmark](https://github.com/technicalpickles/ruby-benchmarks/blob/main/array/map-vs-zip.rb):

```ruby
require 'benchmark/ips'
require 'benchmark/memory'

arrays = {
  small: (1..4).to_a,
  medium: (1..1000).to_a,
  big: (1..50000).to_a,
}

arrays.each do |size, array|
  puts "=== #{size} ==="
  Benchmark.ips do |x|
    x.report(".map { |id| [id] }:") do
      array.map { |id| [id] }
    end

    x.report(".zip:") do
      array.zip
    end

    x.compare! order: :baseline
  end

  Benchmark.memory do |x|
    x.report(".map { |id| [id] }:") do
      array.map { |id| [id] }
    end

    x.report(".zip:") do
      array.zip
    end
  end
end
```

And the results:

```
=== small ===
ruby 3.3.2 (2024-05-30 revision e5a195edf6) [arm64-darwin23]
 .map { |id| [id] }:   405.112k i/100ms
               .zip:   590.068k i/100ms
 .map { |id| [id] }:      4.029M (± 1.1%) i/s  (248.17 ns/i) -     20.256M in   5.027463s
               .zip:      5.846M (± 6.7%) i/s  (171.06 ns/i) -     29.503M in   5.085950s

Comparison:
 .map { |id| [id] }::  4029499.2 i/s
               .zip::  5845845.8 i/s - 1.45x  faster

 .map { |id| [id] }:   240.000  memsize (     0.000  retained)
                         5.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
               .zip:   240.000  memsize (     0.000  retained)
                         5.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
=== medium ===
ruby 3.3.2 (2024-05-30 revision e5a195edf6) [arm64-darwin23]
 .map { |id| [id] }:     2.042k i/100ms
               .zip:     3.533k i/100ms
 .map { |id| [id] }:     20.169k (± 1.7%) i/s   (49.58 μs/i) -    102.100k in   5.063731s
               .zip:     35.106k (± 1.6%) i/s   (28.48 μs/i) -    176.650k in   5.033169s

Comparison:
 .map { |id| [id] }::    20169.1 i/s
               .zip::    35106.3 i/s - 1.74x  faster

 .map { |id| [id] }:    48.040k memsize (     0.000  retained)
                         1.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
               .zip:    48.040k memsize (     0.000  retained)
                         1.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
=== big ===
ruby 3.3.2 (2024-05-30 revision e5a195edf6) [arm64-darwin23]
 .map { |id| [id] }:    44.000 i/100ms
               .zip:    64.000 i/100ms
 .map { |id| [id] }:    425.048 (± 0.7%) i/s    (2.35 ms/i) -      2.156k in   5.072641s
               .zip:    818.393 (± 2.3%) i/s    (1.22 ms/i) -      4.096k in   5.007817s

Comparison:
 .map { |id| [id] }::      425.0 i/s
               .zip::      818.4 i/s - 1.93x  faster

 .map { |id| [id] }:     2.400M memsize (     0.000  retained)
                        50.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
               .zip:     2.400M memsize (     0.000  retained)
                        50.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
```

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Okay, I got something together that covers both the method call and its block

corsonknowles:
Thanks @Earlopain ! 

* Additional specs added, including the derivative cases that you noted
* Both block types/patterns are folded into the `on_send`
* Got the corrector range to match both the method call and the block, but not map's receiver. Also specs demonstrating it works in a method chain.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,73 @@
...
+        def_node_matcher :map_with_array_numblock?, <<~PATTERN
+          (numblock
+            (send _ :map)
+            1
+            (array (lvar _))
+          )
+        PATTERN
+
+        def on_send(node)
+          parent = node.parent
```
Earlopain, line 49:
```suggestion
          return unless (parent = node.parent)
```

There is no parent when no receiver is given, this errors: `map { |id| [id] }`. Unlikely code to exist by itself (unless you maybe write a core extension) but with `ruby-lsp` it runs RuboCop on every keypress. Otherwise I have no complains anymore 👍

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
corsonknowles: Thanks, great catch! I added a spec for this. It also fails with a missing receiver, so I am adding the correct guard clause for that as well.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
corsonknowles: Thanks, great catch! I added a spec for this. It also fails with a missing receiver, so I am adding the correct guard clause for that as well.

corsonknowles:
For the very curious, `.zip` is also faster than `.map! {[_1]}`, and `.each_with_object([]) {|id, object| object << [id]}` is among the slowest of the reasonable strategies.

technicalpickles:
> .each_with_object([]) {|id, object| object << [id]} is among the slowest of the reasonable strategies.

I added it to my benchmark, and it is slower AND takes a **lot**  more memory:

```
=== big ===
ruby 3.3.2 (2024-05-30 revision e5a195edf6) [arm64-darwin23]
 .map { |id| [id] }:    43.000 i/100ms
               .zip:    64.000 i/100ms
.each_with_object([]) {|id, object| object << [id]}:
                        32.000 i/100ms
 .map { |id| [id] }:    458.271 (± 1.3%) i/s    (2.18 ms/i) -      2.322k in   5.067609s
               .zip:    622.430 (± 7.7%) i/s    (1.61 ms/i) -      3.136k in   5.087861s
.each_with_object([]) {|id, object| object << [id]}:
                        327.959 (± 2.4%) i/s    (3.05 ms/i) -      1.664k in   5.077224s

Comparison:
 .map { |id| [id] }::      458.3 i/s
               .zip::      622.4 i/s - 1.36x  faster
.each_with_object([]) {|id, object| object << [id]}::      328.0 i/s - 1.40x  slower

 .map { |id| [id] }:     2.400M memsize (     0.000  retained)
                        50.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
               .zip:     2.400M memsize (     0.000  retained)
                        50.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
.each_with_object([]) {|id, object| object << [id]}:
                         2.454M memsize (     0.000  retained)
                        50.001k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
```

corsonknowles:
@koic Is there anything else I can do to help get this ready to merge?

corsonknowles:
@bbatsov Are you able to take a look at this PR? 

Thanks so much.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,74 @@
...
+        def_node_matcher :map_with_array_numblock?, <<~PATTERN
+          (numblock
+            (send _ :map)
+            1
+            (array (lvar _))
+          )
+        PATTERN
+
+        def on_send(node)
+          return unless (parent = node.parent)
```
rrosenblum, line 49: I'm not opposed to this implementation as you have something that is working. If my memory serves correctly, it tends to be better to start with the parent as the node and extract the children out. I think this would require splitting `on_send` into `on_block` and `on_numblock`.

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
```diff
@@ -0,0 +1,288 @@
...
+  context 'when calling flat_map' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        [1, 2, 3].flat_map {|id| [id]}
+      RUBY
+    end
+  end
+
+  context 'when using Array.wrap' do
+    it 'does not register an offense' do
```
rrosenblum, line 174: It would be cool to add future, or current, support for well known other patterns like this

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
```diff
@@ -0,0 +1,288 @@
...
+      expect_no_offenses(<<~RUBY)
+        [1, 2, 3].select { |id| [id] }
+      RUBY
+    end
+  end
+
+  context 'when making an array of arrays using each_with_object' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        [1,2,3].each_with_object([]) {|id, object| object << [id]}
```
rrosenblum, line 96: It would be cool to add future, or current, support for well known other patterns like this

rrosenblum (commented):
This is a pretty cool optimization for a specific use case. 

I think it would be great to add at least 1 test for a multi-line example

```ruby
[1, 2, 3].map do |id|
  [id]
end
```

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
corsonknowles: That's actually what we started with in this PR. @Earlopain asked to refactor to the `on_send` approach. 

Thank you for this comment though, it really helps contextualize some of the trial and error that went into getting it to work this way. 

I do like the current approach a lot now, it feels very clean to express it in Ruby with guard clauses.

```
module RuboCop
  module Cop
    module Performance
      # Checks for `.map { |id| [id] }` and suggests replacing it with `.zip`.
      
      class UseZipToWrapArrayContents < Base
        extend RuboCop::Cop::AutoCorrector

        MSG = 'Use `.zip` instead of `.map { |id| [id] }`.'

        # @!method map_with_array?(node)
        def_node_matcher :map_with_array?, <<~PATTERN
          (block
            (send _ :map)
            (args (arg _id))
            (array (lvar _id))
          )
        PATTERN

        # Matcher for numblock (shorthand block with _1)
        # @!method map_with_array_numblock?(node)
        def_node_matcher :map_with_array_numblock?, <<~PATTERN
          (numblock
            (send _ :map)
            1
            (array (lvar _))
          )
        PATTERN

        def on_block(node)
          return unless node.method?(:map)
          return unless map_with_array?(node)

          add_offense(node) do |corrector|
            autocorrect(node).call(corrector)
          end
        end

        def on_numblock(node)
          return unless node.method?(:map)
          return unless map_with_array_numblock?(node)

          add_offense(node) do |corrector|
            autocorrect(node).call(corrector)
          end
        end

        def autocorrect(node)
          -> (corrector) do
            corrector.replace(node, correct_replacement(node))
          end
        end

        private

        def correct_replacement(node)
          map_call = node.children.first
          receiver = map_call.receiver.source
          "#{receiver}.zip"
        end
      end
    end
  end
end

```

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Yes! Is this a change request? I added these just so we are exactly specifying the behavior. I do think it can be extended and made to work for this kind of example, as needed by the community.

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Perhaps I should update the description on this one. Array.wrap is a Rails method, so I didn't want to draw too much attention to it. But it doesn't provide equivalent behavior because .wrap is a lot like `[*foo]` meaning when mapped it can take a list of Arrays and non-Arrays and put all the non-arrays into their own array while leaving the Arrays as is.
So I would suspect we would never want to change this, because we'd lose the safe autocorrect aspect of this Cop.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Performance
+      # Checks for `.map { |id| [id] }` and suggests replacing it with `.zip`.
```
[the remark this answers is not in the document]
corsonknowles: I had these and deleted them based on a previous review comment, indicating they were implied. 

The numblock pattern is still very rare, so I'm sympathetic to that.

Kind of makes me which for commit history with a squash and merge approach, but I'm using rebase/amend as it's required here and fairly standard for open source.

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   [1, 2, 3].map { |id| id }
+      #
+      # @example
+      #   # good (no offense)
+      #   [1, 2, 3].map { |id| [id, id] }
+      class UseZipToWrapArrayContents < Base
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = 'Use `.zip` instead of `.map { |id| [id] }`.'
```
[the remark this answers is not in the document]
corsonknowles: Okay, that tuned into a rather large refactor. Done!

corsonknowles:
> This is a pretty cool optimization for a specific use case. 
> 
> 
> 
> I think it would be great to add at least 1 test for a multi-line example
> 
> 
> 
> ```ruby
> 
> [1, 2, 3].map do |id|
> 
>   [id]
> 
> end
> 
> ```

Done!

spec/rubocop/cop/performance/use_zip_to_wrap_array_contents_spec.rb
corsonknowles: Added [*foo] to the specs and made a context for this expected behavior

lib/rubocop/cop/performance/use_zip_to_wrap_array_contents.rb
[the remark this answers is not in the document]
corsonknowles: I read more carefully and I added it as a specific `@bad` example. 

Previous comment: 
https://github.com/rubocop/rubocop-performance/pull/462#discussion_r1760091007

corsonknowles:
@koic @rrosenblum @Earlopain 
Small update here, I renamed this cop in line with:
* https://docs.rubocop.org/rubocop-rspec/development.html#choose-a-name

rrosenblum (commented):
The code looks good to me. You'll need to find a maintainer to actually approve the changes

spec/rubocop/cop/performance/zip_for_array_wrapping_spec.rb
```diff
@@ -0,0 +1,302 @@
...
+    end
+  end
+
+  context 'when using other iterators such as' do
+    context 'when using collect' do
+      it 'does not register an offense as the use of collect is obscure and perhaps intentional' do
+        expect_no_offenses(<<~RUBY)
+          [1, 2, 3].collect { |id| [id] }
+        RUBY
+      end
```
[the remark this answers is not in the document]
corsonknowles: Thanks. Updated to correct on `collect` as well!

corsonknowles:
@bbatsov Hi! Can you help us find a maintainer here to take a look at this contribution? Thank you so much!

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        # Matches numblock form `.map { [_1] }`
+        # @!method map_with_array_numblock?(node)
+        def_node_matcher :map_with_array_numblock?, <<~PATTERN
+          (numblock
+            (send _ :map)
+            1
+            (array (lvar _))
+          )
+        PATTERN
```
dvandersluis, line 49: You can combine these matchers into a single one:

```suggestion
        # @!method map_with_array?(node)
        def_node_matcher :map_with_array?, <<~PATTERN
          {
            (block (send _ RESTRICT_ON_SEND) (args (arg _id)) (array (lvar _id)))
            (numblock (send _ RESTRICT_ON_SEND) 1 (array (lvar _))
            )
          }
        PATTERN
```

Note that also you previously were hardcoding `:map` for `numblock`s which meant `[1, 2, 3].collect { [_1] }` wasn't being detected. I replaced it with `RESTRICT_ON_SEND` above. Please add a test for this case too.

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        def on_send(node)
+          return unless (parent = node.parent)
+          return unless map_with_array?(parent) || map_with_array_numblock?(parent)
```
dvandersluis, line 53:
```suggestion
          return unless map_with_array?(parent)
```

Second half isn't needed after the changes I suggested above.

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,81 @@
...
+        def message(node)
+          format(MSG, original_code: offense_range(node).source.lines.first.chomp)
+        end
+
+        def autocorrect(parent, receiver, corrector)
+          corrector.replace(parent, "#{receiver}.zip")
+        end
+
+        def offense_range(node)
+          @offense_range ||= range_between(node.children.first.loc.selector.begin_pos, node.loc.end.end_pos)
```
dvandersluis, line 76:
```suggestion
          @offense_range ||= node.children.first.loc.selector.join(node.loc.end)
```
Then you can remove `include RangeHelp` too.

dvandersluis:
@corsonknowles sorry that your PR has sat for a bit. I left a couple comments but then I think we can move forward here. Thanks for your patience!

corsonknowles:
> I left a couple comments but then I think we can move forward here. Thanks for your patience!
@dvandersluis 
Awesome, thanks for the simplifications, it's now using 1 matcher method.

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,69 @@
...
+          register_offense(parent, receiver, node)
+        end
+
+        private
+
+        def register_offense(parent, receiver, node)
+          add_offense(offense_range(parent), message: message(node)) do |corrector|
+            autocorrect(parent, receiver, corrector)
+          end
+        end
```
dvandersluis, line 53:
```suggestion
        def register_offense(parent, node)
          offense_range = offense_range(parent)

          add_offense(offense_range(parent), message: message(node)) do |corrector|
            corrector.replace(offense_range, 'zip')
          end
        end
```

Since you're already getting the offense range, we can just replace it with `zip` without needing to replace the whole parent as well. A smaller surface area is preferable to avoid clobbering from other cops' corrections.

This suggestion would also make the `autocorrect` method unnecessary.

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,69 @@
...
+          }
+        PATTERN
+
+        def on_send(node)
+          return unless (parent = node.parent)
+          return unless map_with_array?(parent)
+          return unless (receiver = node.receiver&.source)
+
+          register_offense(parent, receiver, node)
+        end
```
dvandersluis, line 45:
```suggestion
        # @!method map_with_array?(node)
        def_node_matcher :map_with_array?, <<~PATTERN
          {
            (block (call _ RESTRICT_ON_SEND) (args (arg _id)) (array (lvar _id)))
            (numblock (call _ RESTRICT_ON_SEND) 1 (array (lvar _)))
          }
        PATTERN

        def on_send(node)
          return unless (parent = node.parent)
          return unless map_with_array?(parent)
          return unless (receiver = node.receiver&.source)

          register_offense(parent, receiver, node)
        end
        alias on_csend on_send
```

I think we should also have this apply for safe navigation too please (ie. `arr&.map { _1 }`). Could you please add tests for this as well?

dvandersluis:
One additional thing I noticed is that the cop doesn't care about receiver type so `foo.map { _1 }` will be corrected as well. I think this is ok, but can you (1) add tests for this, (2) mark the cop as unsafe and add a `@safety` section to the cop documentation, because we cannot know what type the receiver is, if its `map` is `Enumerable#map` or if it responds to `zip`.

koic:
I'm not entirely convinced about providing this as a built-in cop. Is the code `obj.map { |item| [item] }` a common issue? 

Honestly, if this logic isn't something widely seen as a common use case, I'm hesitant to add a built-in performance cop that specifically targets only the `obj.map { |item| [item] }` logic.

dvandersluis:
I think it's fairly common in Sidekiq for instance, because in order to queue a bulk job, you need an array of arrays of arguments. Often this is used to queue up a job per record in some query for instance, so when there is only a single argument (such as but not limited to `id`), it follows this form. I've also seen `arr.each_slice(1).to_a` which has the same result but is even slower than `map`.

I've seen this pattern with Sidekiq at multiple companies I've worked for, and [here](https://github.com/search?q=repo%3Agitlabhq%2Fgitlabhq+%2Fmap+%5C%7B+%5C%7Cid%5C%7C+%5C%5Bid%5C%5D+%5C%7D%2F&type=code)'s 10 examples at gitlab (not all Sidekiq). In my history, these bulk enqueues can be millions of records.

The pattern is also given in Sidekiq's own wiki page on [Bulk Queueing](https://github.com/sidekiq/sidekiq/wiki/Bulk-Queueing).

koic:
Got it! Thank you for the specific pointers. Now, it would be preferable to come up with a better cop name based on that.

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
corsonknowles: Thanks, I was able to save another allocation after this as well

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
corsonknowles: Implemented!

We had to add this to the node pattern as well to get it to work.

corsonknowles:
> Got it! Thank you for the specific pointers. Now, it would be preferable to come up with a better cop name based on that.

Our internal name before pushing this upstream was `UseZipToWrapArrayContents` 
After reading https://docs.rubocop.org/rubocop-rspec/development.html#choose-a-name I thought `ZipForArrayWrapping` would be consistent with the open source guidance, so I changed it to that.

How about `ZipForArraysOfArrays` or simply `ArrayOfArrays`?

Suggestions welcome!

dvandersluis:
@corsonknowles ahh it's from a change I made upstream in rubocop/rubocop@f450f62461995e4fcac03ed06635bf82c8f7fee7. I'll fix it, thank you.

dvandersluis:
CI is fixed not but "check the oldest supported rubocop version" is failing. This is because in RuboCop 1.48.1, the default target version for tests is 2.6, and numblocks were added in 2.7. Can you please tag your numblock specs with `:ruby27`?

corsonknowles:
> CI is fixed not but "check the oldest supported rubocop version" is failing. This is because in RuboCop 1.48.1, the default target version for tests is 2.6, and numblocks were added in 2.7. Can you please tag your numblock specs with `:ruby27`?

Great catch! Added these tags. I originally added numblock handling because of the InternalAffairs rule. Ruby 2.7 went end of life in Spring '23, so if Rubocop intends to continue to test for and support it much longer, we might want to push some guidance up into the InternalAffairs rule to include this knowledge.

dvandersluis:
> How about ZipForNestedArrays, ZipForArraysOfArrays, or simply NestedArrays? Maybe we still need a verb, like GeneratingNestedArrays ?

@koic do one of these names work for you, or do you have a different suggestion?

koic:
Typically, cop names don't include `For`. In this case, a name without `For` or `Of`, and probably including `Zip`, would be better. I’ll think about it as well. Naming is often the hardest part of implementing a cop.

config/default.yml
```diff
@@ -381,3 +381,9 @@ Performance/UriDefaultParser:
   Description: 'Use `URI::DEFAULT_PARSER` instead of `URI::Parser.new`.'
   Enabled: true
   VersionAdded: '0.50'
+
+Performance/ZipForArrayWrapping:
+  Description: 'Checks for `.map { |id| [id] }` and suggests replacing it with `.zip`.'
```
koic, line 386:
```suggestion
  Description: 'Checks for `map { |id| [id] }` and suggests replacing it with `zip`.'
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Performance
+      # Checks for `.map { |id| [id] }` and suggests replacing it with `.zip`.
```
koic, line 6:
```suggestion
      # Checks for `map { |id| [id] }` and suggests replacing it with `zip`.
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
...
+      #
+      # @example
+      #   # good (no offense)
+      #   [1, 2, 3].map { |id| [id, id] }
+      #
+      # @safety
+      #   This cop is unsafe for novel definitions of `map` and `collect`
+      #   on non-Enumerable objects that do not respond to `zip`.
+      #   To make your object enumerable, define an `each` method
+      #   as described in https://ruby-doc.org/core/Enumerable.html
```
koic, line 30: Can you update to the following?

```suggestion
      # @safety
      #   This cop is unsafe for novel definitions of `map` and `collect`
      #   on non-Enumerable objects that do not respond to `zip`.
      #   To make your object enumerable, define an `each` method
      #   as described in https://ruby-doc.org/core/Enumerable.html
      #
      # @example
      #   # bad
      #   [1, 2, 3].map { |id| [id] }
      #
      #   # bad
      #   [1, 2, 3].map { [_1] }
      #
      #   # good
      #   [1, 2, 3].zip
      #
      #   # good (no offense)
      #   [1, 2, 3].map { |id| id }
      #
      #   # good (no offense)
      #   [1, 2, 3].map { |id| [id, id] }
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
...
+      class ZipForArrayWrapping < Base
+        extend AutoCorrector
+
+        MSG = 'Use `zip` instead of `%<original_code>s`.'
```
koic, line 34: There's no need to pay the price with complex logic, the highlights will suffice this time.

```suggestion
        MSG = 'Use `zip` instead.'
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
...
+        RESTRICT_ON_SEND = Set.new(%i[map collect]).freeze
+
+        # @!method map_with_array?(node)
+        def_node_matcher :map_with_array?, <<~PATTERN
+          {
+            (block ({send csend} _ RESTRICT_ON_SEND) (args (arg _id)) (array (lvar _id)))
+            (numblock ({send csend} _ RESTRICT_ON_SEND) 1 (array (lvar _)))
```
koic, line 41:
```suggestion
            (block (call !nil? RESTRICT_ON_SEND) (args (arg _)) (array (lvar _)))
            (numblock (call !nil? RESTRICT_ON_SEND) 1 (array (lvar _)))
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
...
+          }
+        PATTERN
+
+        def on_send(node)
+          return unless map_with_array?(node.parent)
+          return if node.receiver.nil?
```
koic, line 47: This is not necessary in the above pattern.

```suggestion
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
...
+        def register_offense(node)
+          offense_range = offense_range(node.parent)
+          add_offense(offense_range, message: message(node)) do |corrector|
+            corrector.replace(offense_range, 'zip')
+          end
+        end
+
+        def message(node)
+          format(MSG, original_code: offense_range(node).source.lines.first.chomp)
+        end
```
koic, line 64:
```suggestion
        def register_offense(node)
          message = format(MSG, current: offense_range.source)

          add_offense(offense_range, message: message) do |corrector|
            corrector.replace(offense_range, 'zip')
          end
        end
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        def offense_range(node)
+          @offense_range ||= node.children.first.loc.selector.join(node.loc.end)
```
koic, line 67:
```suggestion
          node.send_node.loc.selector.join(node.loc.end)
```

spec/rubocop/cop/performance/zip_for_array_wrapping_spec.rb
```diff
@@ -0,0 +1,389 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Performance::ZipForArrayWrapping, :config do
+  context 'when using map with array literal' do
+    it 'registers an offense and corrects to use zip with no arguments' do
+      expect_offense(<<~RUBY)
+        [1, 2, 3].map { |id| [id] }
+                  ^^^^^^^^^^^^^^^^^ Use `zip` instead of `map { |id| [id] }`.
```
koic, line 8: Can you update the message in this test?

```suggestion
                  ^^^^^^^^^^^^^^^^^ Use `zip` instead.
```

lib/rubocop/cop/performance/zip_for_array_wrapping.rb
corsonknowles: It's all so elegant and simple now!

spec/rubocop/cop/performance/zip_for_array_wrapping_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+          [1, 2, 3].collect { |id| [id] }
+                    ^^^^^^^^^^^^^^^^^^^^^ Use `zip` without a block argument instead.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          [1, 2, 3].zip
+        RUBY
+      end
+
+      it 'registers an offense for collect with a numblock, :ruby27' do
```
dvandersluis, line 156:
```suggestion
      it 'registers an offense for collect with a numblock', :ruby27 do
```

corsonknowles:
> Typically, cop names don't include `For`. In this case, a name without `For` or `Of`, and probably including `Zip`, would be better. I’ll think about it as well. Naming is often the hardest part of implementing a cop.

I think you may have just named it perfectly in the revised Description & Message. If we focus on what the fix is, like we would suggest in code review, the performant choice is to use "Zip Without a Block" -- I renamed it, let me know what you think.

lib/rubocop/cop/performance/zip_without_block.rb
```diff
@@ -0,0 +1,65 @@
...
+      #   [1, 2, 3].map { [_1] }
+      #
+      #   # good
+      #   [1, 2, 3].zip
+      #
+      #   # good (no offense)
+      #   [1, 2, 3].map { |id| id }
+      #
+      #   # good (no offense)
+      #   [1, 2, 3].map { |id| [id, id] }
```
koic, line 28: I think the following example is neither lacking nor excessive for user :-)

```suggestion
      # @example
      #   # bad
      #   [1, 2, 3].map { |id| [id] }
      #
      #   # good
      #   [1, 2, 3].zip
```

koic:
`Performance/ZipWithoutBlock` cop is a better name than `Performance/ZipForArrayWrapping` cop. I also considered the name `Performance/MapAsArrayWrapper` cop, but I’m not confident about it. If no better name comes to mind for a while, `Performance/ZipWithoutBlock` cop will likely be the candidate.

koic:
Since no particularly better name has been suggested afterward, let's proceed with this. Thank you.

PR #463, opened 2024-09-06
[Fix #454] Fix false positives for `Performance/BigDecimalWithNumericArgument`

### LINKED ISSUE #454: Correcting `Performance/BigDecimalWithNumericArgument` offences leads to slower code
`Performance/BigDecimalWithNumericArgument` suggests that `BigDecimal(2000)` is more performant as `BigDecimal("2000")` but that is not the case for recent versions of Ruby/BigDecimal:

```ruby
require 'benchmark/ips'
require 'bigdecimal'

Benchmark.ips do |x|
  x.report('integer string new')  { BigDecimal("2000") }
  x.report('integer new')         { BigDecimal(2000) }
  x.compare!
end
```

Gives for me (using Ruby 3.3.3 and BigDecimal 3.1.8):
```
ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [x86_64-darwin22]
  integer string new   228.898k i/100ms
         integer new   337.726k i/100ms
  integer string new      2.520M (± 5.9%) i/s -     12.589M in   5.015809s
         integer new      4.443M (±10.5%) i/s -     21.952M in   5.006675s

Comparison:
         integer new:  4442710.0 i/s
  integer string new:  2519800.0 i/s - 1.76x  slower
```
Having dug a little bit through various Ruby/BigDecimal versions, I think the improvement was introduced in BigDecimal 3.1.0 (which was first included in Ruby 3.1.0), I suspect this PR led to the performance improvement https://github.com/ruby/bigdecimal/pull/178

## Expected behavior

There should be no offence for the faster code `BigDecimal(2000)`

## Actual behavior

There was an offence. 

Run rubocop-performance against a file containing `BigDecimal(2000)`

## RuboCop version

```
1.61.0 (using Parser 3.3.3.0, rubocop-ast 1.31.3, running on ruby 3.3.3) [x86_64-darwin22]
  - rubocop-performance 1.21.1
```

koic:
Fixes #454.

This PR fixes false positives for `Performance/BigDecimalWithNumericArgument` when using BigDecimal 3.1+.

The bad and good examples for this cop are reversed between BigDecimal 3.0 and 3.1.

Since BigDecimal 3.1 is the default gem in Ruby 3.1, this PR reverses the bad and good examples when targeting Ruby 3.1+. So, the goal of this PR is to address many cases where the meaning was entirely reversed.

To avoid introducing excessive complexity related to version dependencies, detection is disabled for Ruby 3.0 and below.

Ideally, a solution that prioritizes the BigDecimal version would be best in the future, but this PR does not take that into account.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/performance/big_decimal_with_numeric_argument.rb
+++ b/lib/rubocop/cop/performance/big_decimal_with_numeric_argument.rb
@@ -3,45 +3,48 @@
 module RuboCop
   module Cop
     module Performance
-      # Identifies places where numeric argument to BigDecimal should be
-      # converted to string. Initializing from String is faster
-      # than from Numeric for BigDecimal.
+      # Identifies places where string argument to `BigDecimal` should be
+      # converted to numeric. Initializing from Integer is faster
+      # than from String for BigDecimal.
       #
       # @example
       #   # bad
-      #   BigDecimal(1, 2)
-      #   4.to_d(6)
-      #   BigDecimal(1.2, 3, exception: true)
-      #   4.5.to_d(6, exception: true)
-      #
-      #   # good
       #   BigDecimal('1', 2)
       #   BigDecimal('4', 6)
       #   BigDecimal('1.2', 3, exception: true)
       #   BigDecimal('4.5', 6, exception: true)
       #
+      #   # good
+      #   BigDecimal(1, 2)
+      #   4.to_d(6)
+      #   BigDecimal(1.2, 3, exception: true)
+      #   4.5.to_d(6, exception: true)
+      #
       class BigDecimalWithNumericArgument < Base
         extend AutoCorrector
+        extend TargetRubyVersion
 
-        MSG = 'Convert numeric literal to string and pass it to `BigDecimal`.'
+        minimum_target_ruby_version 3.1
+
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/performance/big_decimal_with_numeric_argument.rb
```diff
@@ -3,45 +3,48 @@
...
+      #   BigDecimal(1, 2)
+      #   4.to_d(6)
+      #   BigDecimal(1.2, 3, exception: true)
+      #   4.5.to_d(6, exception: true)
+      #
       class BigDecimalWithNumericArgument < Base
         extend AutoCorrector
+        extend TargetRubyVersion
+
+        minimum_target_ruby_version 3.1
```
Earlopain, line 27: Can you use `requires_gem` API here? bigdecimal is becoming a bundled gem so many already have it in their lockfile and the ruby version is barely relevant anymore. This would stop this cop from doing something if the user doesn't use bundler but I think this is the way to go

Earlopain: Ah, right. You already meant `requires_gem` wit with that

PR #466, opened 2024-09-13
[Fix #385] Disable `Performance/BlockGivenWithExplicitBlock` by default

### LINKED ISSUE #385: Suggestion from Performance/BlockGivenWithExplicitBlock is often slower
`Performance/BlockGivenWithExplicitBlock` provides bad guidance and should be removed. The benchmark which was given when it was introduced didn't test the case that a block was given, when that happens `block` is significantly slower than the other option.

For Ruby 3.2+ I made [`if block` faster](https://github.com/ruby/ruby/pull/6286), but prior to that the advice from this cop was always wrong. Even in 3.2+ I believe it's bad advice as there's only a very narrow usage for which `block` is (very slightly) faster, and even in those cases it's likely a future refactor the user will accidentally introduce the slow case (ex. moving from `if block` to `if block && something_else`)

``` ruby
require 'bundler/inline'

gemfile(true) do
  gem 'benchmark-ips'
end

def if_block(&block)
  !!block
end

def if_block_given(&block)
  !!block_given?
end

Benchmark.ips do |x|
  x.report('block')                  { if_block }
  x.report('block w/ block')         { if_block {} }
  x.report('block_given?')           { if_block_given }
  x.report('block_given? w/ block')  { if_block_given {} }
  x.compare!
end
```

```
            block     1.612M i/100ms
   block w/ block   508.391k i/100ms
     block_given?     1.309M i/100ms
block_given? w/ block
                         1.068M i/100ms
            block     15.804M (± 1.8%) i/s -     79.000M in   5.000346s
   block w/ block      5.017M (± 0.5%) i/s -     25.420M in   5.067106s
     block_given?     13.297M (± 1.7%) i/s -     66.746M in   5.021282s
block_given? w/ block
                         10.745M (± 0.6%) i/s -     54.450M in   5.067787s

Comparison:
            block: 15804072.8 i/s
     block_given?: 13296844.8 i/s - 1.19x  slower
block_given? w/ block: 10744860.7 i/s - 1.47x  slower
   block w/ block:  5016692.5 i/s - 3.15x  slower
```

Earlopain:
See linked issue for benchmark details.

I tested with YJIT and its even worse:
```
ruby 3.4.0dev (2024-09-05T09:43:46Z master 63cbe3f6ac) +YJIT [x86_64-linux]
               block     1.635M i/100ms
      block w/ block   285.930k i/100ms
        block_given?     1.867M i/100ms
block_given? w/ block
                         1.497M i/100ms
               block     22.611M (±12.7%) i/s   (44.23 ns/i) -    111.180M in   5.033721s
      block w/ block      3.034M (± 8.5%) i/s  (329.59 ns/i) -     15.154M in   5.032609s
        block_given?     24.935M (±10.7%) i/s   (40.10 ns/i) -    123.254M in   5.002983s
block_given? w/ block
                         23.007M (±14.7%) i/s   (43.47 ns/i) -    113.782M in   5.058213s

Comparison:
        block_given?: 24935388.3 i/s
block_given? w/ block: 23006920.5 i/s - same-ish: difference falls within error
               block: 22610939.8 i/s - same-ish: difference falls within error
      block w/ block:  3034047.2 i/s - 8.22x  slower
```

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_disable_block_given.md
+++ b/changelog/change_disable_block_given.md
@@ -1,0 +1,1 @@
+* [#385](https://github.com/rubocop/rubocop-performance/issues/385): Disable `Performance/BlockGivenWithExplicitBlock` by default. ([@earlopain][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -32,8 +32,11 @@ Performance/BindCall:
 
 Performance/BlockGivenWithExplicitBlock:
   Description: 'Check block argument explicitly instead of using `block_given?`.'
-  Enabled: pending
+  # This cop was created due to a mistake in microbenchmark.
+  # https://github.com/rubocop/rubocop-performance/issues/385
+  Enabled: false
   VersionAdded: '1.9'
+  VersionChanged: <<next>>
 
 Performance/Caller:
   Description: >-
--- a/lib/rubocop/cop/performance/block_given_with_explicit_block.rb
+++ b/lib/rubocop/cop/performance/block_given_with_explicit_block.rb
@@ -6,6 +6,9 @@ module RuboCop
       # Identifies unnecessary use of a `block_given?` where explicit check
       # of block argument would suffice.
       #
+      # NOTE: This cop produces code with significantly worse performance when a
+      # block is being passed to the method and as such should not be enabled.
+      #
       # @example
       #   # bad
       #   def method(&block)


koic:
If the speed is reversed, shouldn't the behavior be updated to reverse the bad and good cases instead of disabling the cop?

Earlopain:
I would rather not make that decision. The difference today is smaller than it was when the cop was first introduced and I find it likely to be further optimized in Ruby itself in the future, where the cop can can then again produce worse results.

The results are also only consistently in favor of `block_given?` when YJIT is enabled, otherwise `if block` (without actually passing in a block) is still the fastest variant):
```
               block:  9336250.8 i/s
        block_given?:  8365028.9 i/s - same-ish: difference falls within error
block_given? w/ block:  7620740.6 i/s - 1.23x  slower
      block w/ block:  2334882.5 i/s - 4.00x  slower
```

Overall "it depends". Per https://github.com/rubocop/rubocop-performance/issues/385#issuecomment-1819638731 there is also a difference if the block parameter is defined in the method signature.

koic:
> The difference today is smaller than it was when the cop was first introduced and I find it likely to be further optimized in Ruby itself in the future

I see. That would make sense.

PR #474, opened 2024-10-14
Add new `Performance/StringBytesize` cop

viralpraxis:
-----------------

* [x] Added tests.


lib/rubocop/cop/performance/string_bytesize.rb
```diff
@@ -0,0 +1,49 @@
...
+          (call (call !nil? :bytes) {:size :length :count})
+        MATCHER
+
+        def on_send(node)
+          string_bytes_method?(node) do
+            add_offense(node) do |corrector|
+              corrector.replace(node, replacement(node))
+            end
+          end
+        end
```
dvandersluis, line 38: Using a range will target the offense directly to `bytes.size` (etc.) without including the receiver. It'll also make the correction simpler (and you won't need your `replacement` method anymore). Tests will need to be updated to refine the `^^^^^`s to the smaller area of course.

```suggestion
        def on_send(node)
          string_bytes_method?(node) do
            range = node.receiver.loc.selector.begin.join(node.loc.expression.end)

            add_offense(range) do |corrector|
              corrector.replace(range, 'bytesize')
            end
          end
        end
```

lib/rubocop/cop/performance/string_bytesize.rb
```diff
@@ -0,0 +1,49 @@
...
+      #   string_var.bytesize
+      #   "foobar".bytesize
+      class StringBytesize < Base
+        extend AutoCorrector
+
+        MSG = 'Use `String#bytesize` instead of calculating the size of the bytes array.'
+        RESTRICT_ON_SEND = %i[size length count].freeze
+
+        def_node_matcher :string_bytes_method?, <<~MATCHER
+          (call (call !nil? :bytes) {:size :length :count})
```
dvandersluis, line 29: Can you add a test for `bytes.size` with no receiver?

viralpraxis: There's already a test for this case: https://github.com/rubocop/rubocop-performance/pull/474/files#diff-6bd9463058468a2a8ab5cd49bd6023946d72b460a0382226c7d16649a5db1421R78-R82

lib/rubocop/cop/performance/string_bytesize.rb
viralpraxis: Thanks, done.

lib/rubocop/cop/performance/string_bytesize.rb
dvandersluis: My bad, missed it. Thanks!

koic:
I understand the reasoning, but in Rails, replacing `42.bytes.size` with `42.bytesize` results in `NoMethodError`. This makes me concerned about its usefulness compared to the risk of false positives.

viralpraxis:
Yes, that's an issue. But isn't `Integer#size` less common than `Array#size`? Could we simply disable autocorrection, or only check for the `length` and `count` methods?

viralpraxis:
Any thoughts? @koic @dvandersluis

I completely understand that we should avoid false positives (if possible), but in this particular case I believe both true and false positives are quite rate; at the same time unnecessary array allocations are painful in case of huge strings. I was inspired by code like this

```ruby
class BytesizeValidator < ActiveModel::Validations::LengthValidator
  def validate_each(record, attribute, value)
    super(record, attribute, value&.bytes)
  end
end
```

which was fixed by me to

```ruby
class BytesizeValidator < ActiveModel::Validations::LengthValidator
  def validate_each(record, attribute, value)
    super(record, attribute, value&.dup&.force_encoding(Encoding::BINARY))
  end
end
```

to reduce memory consumption

dvandersluis:
I don't think `42.bytes.size` is very likely in a Rails app (because it's essentially calling `Integer#size`), but it could be avoided in most cases by having this cop not report if the receiver is a numeric literal. It won't cover 100% of cases (ie. variables/methods), but would probably cover most.

It'd be good to have a reference and/or benchmarks for this performance suggestion. Your example doesn't actually make a good case for changing `bytes.size` to `bytesize` on its own, just avoiding using the bytes array (but that would apply to any unnecessary use of an array, not specifically to `bytes.size`).

> Could we simply disable autocorrection, or only check for the length and count methods?

I wouldn't change this to only check for `length` and `count`, but that wouldn't solve the issue anyways, since all three are used in many contexts. I think that'd just diminish the utility of this cop. BTW - If this PR was to be accepted it'd be ideal for the tests to include all three methods (size/length/count).

koic:
@dvandersluis's opinion looks good. @viralpraxis Can you proceed with this?
However, for safety, it should use `Safe: false` instead of `SafeAutoCorrect: false` due to the risk of false positives.

viralpraxis:
@koic I agree with @dvandersluis too. I will apply the necessary changes.

koic:
Thanks!

PR #476, opened 2024-11-02
Caller: use lonely operator to safely call `#first`

chastell:
The Caller cop currently replaces `caller[42]` with `caller(43..43).first`.

If the caller stack is short, the former returns `nil` while the latter raises a `NoMethodError` (_undefined method `first' for nil_).

This PR fixes this by using the lonely operator (`&.`) to call `#first` only if the result is non-`nil`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_caller_use_lonely_operator_to_safely.md
+++ b/changelog/fix_caller_use_lonely_operator_to_safely.md
@@ -1,0 +1,1 @@
+* [#476](https://github.com/rubocop/rubocop-performance/pull/476): Caller: use lonely operator to safely call `#first`. ([@chastell][])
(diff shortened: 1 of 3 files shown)


koic:
Thank you for the suggestion. I understand the reasoning, but this solution cannot be adopted for two reasons:

- Safe navigation is not always necessary. In cases where it is certainly unnecessary, it becomes redundant.  
- In situations where an error is preferable to incorrect behavior, this approach would result in unintended behavior. Unexpected behavior is more problematic than an explicit error.  

In other words, while safe navigation is appropriate in the context of this PR, there are cases where it is not.

PR #480, opened 2024-12-18
Fix `Performance/Squeeze` cop error on frozen AST string node value

viralpraxis:
Since sometimes AST string nodes might have their values frozen, we should call `dup` explicitly to make sure `to_string_literal` (which does not work with frozen strings) does not raise. Maybe we should also apply a patch on RuboCop's core side to fix the `to_string_literal` itself (it now uses `force_encoding` on its argument).

See [1] and [2] for details.

[1] https://github.com/ruby/prism/issues/3309
[2] https://github.com/rubocop/rubocop-ast/issues/342

-----------------

* [X] Added tests.

### DIFF
--- a/lib/rubocop/cop/performance/squeeze.rb
+++ b/lib/rubocop/cop/performance/squeeze.rb
@@ -46,7 +46,11 @@ module RuboCop
             message = format(MSG, current: bad_method, prefer: good_method)
 
             add_offense(node.loc.selector, message: message) do |corrector|
-              string_literal = to_string_literal(replace_str)
+              # FIXME: When requiring only RuboCop 1.70.0 and above,
+              # `dup` in `replace_str.dup` becomes unnecessary, as
+              # frozen strings are handled in the `to_string_literal`
+              # implementation. Please remove it.
+              string_literal = to_string_literal(replace_str.dup)
               new_code = "#{receiver.source}#{node.loc.dot.source}#{good_method}(#{string_literal})"
 
               corrector.replace(node, new_code)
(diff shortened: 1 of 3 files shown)


Earlopain:
Yes, I think this should instead be handled in rubocop itself instead of potentially having to handle this at every callsite. https://github.com/rubocop/rubocop/blob/9f2ffa2270bbc174b2a622350d120b05672a1ce5/lib/rubocop/cop/util.rb#L198-L201. Other `force_encoding` in rubocop are already preceeded with a `dup` (and there's special handling for jruby for some reason)

viralpraxis:
> Yes, I think this should instead be handled in rubocop itself instead of potentially having to handle this at every callsite. https://github.com/rubocop/rubocop/blob/9f2ffa2270bbc174b2a622350d120b05672a1ce5/lib/rubocop/cop/util.rb#L198-L201. Other `force_encoding` in rubocop are already preceeded with a `dup` (and there's special handling for jruby for some reason)

Shouldn't it still be fixed here since latest `rubocop-performance` might work with old `rubocop`'?

Earlopain:
I guess, but for this I would just be tempted to force those select few users that run into this to upgrade instead. It is so very specific and doesn't look like something that someone would actually write (?). Not my decision though.

viralpraxis:
> doesn't look like something that someone would actually write

That's actually the code I wrote right after learning about "character literals" 😄  

Anyway, I'll open a PR at rubocop's side to address this issue

lib/rubocop/cop/performance/squeeze.rb
```diff
@@ -46,7 +46,7 @@ def on_send(node)
             message = format(MSG, current: bad_method, prefer: good_method)
 
             add_offense(node.loc.selector, message: message) do |corrector|
-              string_literal = to_string_literal(replace_str)
+              string_literal = to_string_literal(replace_str.dup)
```
koic, line 49: Can you add a FIXME comment? e.g.:

```suggestion
              # FIXME: When requiring RuboCop 1.70.0 and above,
              # `dup` in `replace_str.dup` becomes unnecessary, so please remove it.
              string_literal = to_string_literal(replace_str.dup)
```

koic:
Thanks!

PR #488, opened 2025-01-25
Change `Performance/CollectionLiteralInLoop` to not register offenses for `Array#include?` that are optimized directly in Ruby.

### LINKED ISSUE #482: Investigate `Performance/CollectionLiteralInLoop` for Ruby 3.4
Ruby 3.4 added optimizations which I believe are relevant to this cop. See https://github.com/ruby/ruby/pull/12123

Basically specialized instruction handling for `[:a, :b].include?(@ivar)`. I haven't benchmarked yet but maybe this cop can be disabled on ruby 3.4 for most common simple cases if numers are good. Will investigate later.

Earlopain:
Fix #482. Since `include?` on arrays are the only instances I ever run across this cop, I think it makes sense to add special handling for this.

On Ruby 3.4 this does no array allocations:
```rb
require "memory_profiler"

class Foo
  def bar
    Bar.new
  end
end

class Bar
  def baz
  end
end

foo = Foo.new
report = MemoryProfiler.report do
  [1,2].include?(foo.bar.baz)
end.pretty_print
```

Also, this code is simply slower on Ruby 3.4:

```rb
require "benchmark/ips"

local = [301, 302]
val = 301
Benchmark.ips do |x|
  x.report('local') do
    local.include?(val)
  end

  x.report('literal') do
    [301, 302].include?(val)
  end
  x.compare!
end
```

```
$ ruby test.rb
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               local     1.822M i/100ms
             literal     2.296M i/100ms
               local     18.235M (± 1.6%) i/s   (54.84 ns/i) -     92.906M in   5.096277s
             literal     22.807M (± 1.5%) i/s   (43.85 ns/i) -    114.789M in   5.034289s

Comparison:
             literal: 22806932.0 i/s
               local: 18235340.7 i/s - 1.25x  slower
```

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/performance/collection_literal_in_loop_spec.rb
+++ b/spec/rubocop/cop/performance/collection_literal_in_loop_spec.rb
@@ -269,4 +269,71 @@ RSpec.describe RuboCop::Cop::Performance::CollectionLiteralInLoop, :config do
       RUBY
     end
   end
+
+  context 'when Ruby >= 3.4', :ruby34 do
+    it 'registers an offense for `include?` on a Hash literal' do
+      expect_offense(<<~RUBY)
+        each do
+          { foo: :bar }.include?(:foo)
+          ^^^^^^^^^^^^^ Avoid immutable Hash literals in loops. It is better to extract it into a local variable or a constant.
+        end
+      RUBY
+    end
+
+    it 'registers an offense for other array methods' do
+      expect_offense(<<~RUBY)
+        each do
+          [1, 2, 3].index(foo)
+          ^^^^^^^^^ Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
+        end
+      RUBY
+    end
+
+    context 'when using an Array literal and calling `include?`' do
+      [
+        '"string"',
+        'self',
+        'local_variable',
+        'method_call',
+        '@instance_variable'
+      ].each do |argument|
+        it "registers no offense when the argument is #{argument}" do
+          expect_no_offenses(<<~RUBY)
+            #{'local_variable = 123' if argument == 'local_variable'}
+            array.all? do |e|
+              [1, 2, 3].include?(#{argument})
+            end
+          RUBY
+        end
+
+        it "registers no offense when the argument is #{argument} with method chain" do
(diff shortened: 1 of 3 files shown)


Earlopain:
:thinking: how do I get the oldest rubocop version tests to pass when it doesn't know about `ruby34`?

Earlopain:
I just added it manually for now, can be removed when the minimum supported rubocop gets bumped.

spec/rubocop/cop/performance/collection_literal_in_loop_spec.rb
```diff
@@ -269,4 +269,74 @@
...
+      [
+        '"string"',
+        'self',
+        'local_variable',
+        'method_call',
+        '@instance_variable'
+      ].each do |argument|
+        it "registers no offense when the argument is #{argument}" do
+          expect_no_offenses(<<~RUBY)
+            #{'local_variable = 123' if argument == 'local_variable'}
```
koic, line 305: The presence or absence of this string interpolation does not change the testing result, but is this code meaningful in relation to the test's behavior?

Earlopain: Removing it makes it send instead of lvar, but both are handled so the test doesn't visibly change

spec/rubocop/cop/performance/collection_literal_in_loop_spec.rb
```diff
@@ -269,4 +269,74 @@
       RUBY
     end
   end
+
+  context 'when Ruby >= 3.4', :ruby34 do
+    # TODO: Remove when the minimum supported RuboCop is 1.62.0
+    let(:ruby_version) { 3.4 }
```
koic, line 275: The current latest version of RuboCop Performance already requires RuboCop 1.72+. Can you remove it?

Earlopain: Nice, I removed it

koic:
Thanks!

PR #498, opened 2025-05-03
Fix `Performance/Count` cop error on empty selector block

viralpraxis:
Something like

```ruby
collection.reject { }.size
```

would lead to an error

-----------------

* [X] Added tests.

### DIFF
--- a/changelog/fix_performance_count_cop_error_on_empty_selector_block.md
+++ b/changelog/fix_performance_count_cop_error_on_empty_selector_block.md
@@ -1,0 +1,1 @@
+* [#498](https://github.com/rubocop/rubocop-performance/pull/498): Fix `Performance/Count` cop error on empty selector block. ([@viralpraxis][])
--- a/lib/rubocop/cop/performance/count.rb
+++ b/lib/rubocop/cop/performance/count.rb
@@ -87,7 +87,9 @@ module RuboCop
         end
 
         def eligible_node?(node)
-          !(node.parent && node.parent.block_type?)
+          return false if node.parent&.block_type?
+
+          node.receiver.call_type? || node.receiver.body
         end
 
         def source_starting_at(node)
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/performance/count_spec.rb
```diff
@@ -353,4 +353,17 @@ def count(&block)
       RUBY
     end
   end
+
+  context 'with `reject` with empty block body' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        array.reject {}.size
+              ^^^^^^^^^^^^^^ Use `count` instead of `reject...size`.
```
koic, line 361: Since empty block is not typically used, and the code produced by autocorrection would likely be non-idiomatic, so this case should probably be ignored by `expect_no_offenses`.

viralpraxis: Makes sense -- fixed.

lib/rubocop/cop/performance/count.rb
```diff
@@ -87,7 +87,7 @@ def autocorrect(corrector, node, selector_node, selector)
         end
 
         def eligible_node?(node)
-          !(node.parent && node.parent.block_type?)
+          !(node.parent && node.parent.block_type?) && (node.receiver.call_type? || node.receiver.body)
```
koic, line 90: The condition seems a bit complex. How about refactoring it as follows?

```suggestion
          return false if node.parent&.block_type?

          node.receiver.call_type? || node.receiver.body
```

changelog/fix_performance_count_cop_error_on_empty_selector_block.md
```diff
@@ -0,0 +1 @@
+* [#497](https://github.com/rubocop/rubocop-performance/pull/497): Fix `Performance/Count` cop error on empty selector block. ([@viralpraxis][])
```
koic, line 1:
```suggestion
* [#498](https://github.com/rubocop/rubocop-performance/pull/498): Fix `Performance/Count` cop error on empty selector block. ([@viralpraxis][])
```

PR #499, opened 2025-05-04
Add new `Performance/ConditionalDefinition` cop

viralpraxis:
There's a well-known pattern of defining methods conditionally, especially in gems:

```ruby
def foo
  if RUBY_VERSION < '4.0.'
    ...
  else
    ...
  end
end
```

If `foo` is a hotspot, its performance can be improved (see benchmarks below) by moving the conditional logic outside the method body definition:

```ruby
if RUBY_VERSION < '4.0.'
  def foo
    ...
  end
else
  def foo
    ...
  end
end
```

I'd like to propose a new `Performance/ConditionalDefinition` cop that detects conditional method definitions, as described above. We can start by matching constant expressions such as `{RUBY_VERSION,RUBY_RELEASE_DATE,...} <op> ...`, and generalize the detection logic later.

```ruby

require 'benchmark/ips'

class ConditionInBody
  def foo
    if RUBY_VERSION >= '3.2'
      1
    else
      2
    end
  end
end

class ConditionalDefinition
  if RUBY_VERSION >= '3.2'
    def foo
      1
    end
  else
    def foo
      2
    end
  end
end

obj1 = ConditionInBody.new
obj2 = ConditionalDefinition.new

Benchmark.ips do |x|
  x.report('condition in body') do
    obj1.foo
  end

  x.report('conditional method definition') do
    obj2.foo
  end

  x.compare!
end
```

```bash
$ ruby bm.rb
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [x86_64-linux]
   condition in body     1.170M i/100ms
conditional method definition
                         2.228M i/100ms
   condition in body     11.683M (± 2.6%) i/s   (85.60 ns/i) -     58.485M in   5.009850s
conditional method definition
                         22.148M (± 1.0%) i/s   (45.15 ns/i) -    111.418M in   5.031059s

Comparison:
conditional method definition: 22148111.6 i/s
   condition in body: 11682500.0 i/s - 1.90x  slower
```

I'm not an expert in YJIT (or any other MRI JIT compiler), but AFAIK even when YJIT determines that `RUBY_VERSION <op> ...` is constant, it still needs to check a guard clause for every method call. So there's still a performance improvement:

```bash
$ ruby --yjit bm.rb
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +YJIT +PRISM [x86_64-linux]
   condition in body     1.853M i/100ms
conditional method definition
                         3.356M i/100ms
   condition in body     22.752M (± 0.4%) i/s   (43.95 ns/i) -    114.907M in   5.050444s
conditional method definition
                         51.223M (± 0.4%) i/s   (19.52 ns/i) -    258.439M in   5.045467s

Comparison:
conditional method definition: 51223091.5 i/s
   condition in body: 22752233.2 i/s - 2.25x  slower
```

**NOTE**: This is a WIP, and the cop is not yet ready -- at this point I just want to get feedback on the idea so I can continue working on it.

-----------------

* [X] Added tests.


viralpraxis:
Could someone confirm whether this cop makes sense? If so I’ll continue working on it

PR #504, opened 2025-06-13
Fix autocorrection syntax error for `Performance/Count` with multiline calls

lovro-bikic:
Currently, `Performance/Count` will autocorrect:
```ruby
foo
  .select(&:something)
  .count
  .positive?
```
to:
```ruby
foo
  .count(&:something)

  .positive?
```
which is not valid syntax. This PR fixes autocorrection to:
```ruby
foo
  .count(&:something)
  .positive?
```

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/performance/count_spec.rb
+++ b/spec/rubocop/cop/performance/count_spec.rb
@@ -1,12 +1,22 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Performance::Count, :config do
-  shared_examples 'selectors' do |selector|
+  shared_examples 'selectors' do |selector, negate_condition|
     it "registers an offense for using array.#{selector}...size" do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/performance/count_spec.rb
```diff
@@ -1,54 +1,124 @@
...
+      if negate_condition
+        expect_correction(<<~RUBY)
+          [1, 2, 3].count { |e| !(e.even?) }
+        RUBY
+      else
+        expect_correction(<<~RUBY)
+          [1, 2, 3].count { |e| e.even? }
+        RUBY
+      end
     end
```
lovro-bikic, line 20: I took the liberty of adding correction expectations for all examples which were missing it; let me know if I should move this to a separate PR.

PR #529, opened 2026-07-30
[Fix #438] Fix an incorrect autocorrect for Performance/ConstantRegex…

### LINKED ISSUE #438: Autocorrection for Performance/ConstantRegexp is not working with pattern matching
## Expected behavior

The autocorrection for `Performance/ConstantRegexp` should fix the offense even with pattern matching.

## Actual behavior

The autocorrection for `Performance/ConstantRegexp` prepends the end delimiter `/o`, leading to the following error:

```ruby
NODE_IN: unknown node (NODE_ONCE)
```

I have a method that uses pattern matching:

```ruby
# constants/errors.rb
UNINITIALIZED_CONSTANT = "uninitialized constant".freeze
UNDEFINED_VARIABLE = "undefined local variable".freeze
    
def call
  case exception.message
  in /#{Constants::Errors::UNINITIALIZED_CONSTANT}/
    # do something
  in /#{Constants::Errors::UNDEFINED_VARIABLE}/
    # do something
  else
    # do something
  end
end
```

It works well but `rubocop-performance` raises the following offense:

```ruby
Performance/ConstantRegexp: Extract this regexp into a constant, memoize it, or append an `/o` option to its options
```

The autocorrect prepends `/o`

```ruby
# constants/errors.rb
UNINITIALIZED_CONSTANT = "uninitialized constant".freeze
UNDEFINED_VARIABLE = "undefined local variable".freeze

def call
  case exception.message
  in /#{Constants::Errors::UNINITIALIZED_CONSTANT}/o
    # do something
  in /#{Constants::Errors::UNDEFINED_VARIABLE}/o
    # do something
  else
    # do something
  end
end
```

and it breaks the code with the following error:

```ruby
NODE_IN: unknown node (NODE_ONCE)
```

FYI I tried an `if/else` version and it works correctly

```ruby
def call
  if exception.message.match?(/#{Constants::Errors::UNINITIALIZED_CONSTANT}/o)
    # do something
  elsif exception.message.match?(/#{Constants::Errors::UNDEFINED_VARIABLE}/o)
    # do something
  else
    # do something
  end
end
 ```

## RuboCop version

```
rubocop (1.57.2)
rubocop-performance (1.19.1)
```

Note: I use the `standard` gem

pcbeingused333:
…p in pattern matching

Fixes #438.

`Performance/ConstantRegexp` autocorrects an interpolated-constant regexp by appending the `/o` option. When the regexp is used as a pattern in pattern matching (`case/in`, one-line `in`), `/o` is invalid Ruby and the autocorrection produces broken code (`NODE_ONCE: unknown node`).

This skips registering an offense when the regexp is in a pattern-matching pattern position, mirroring the existing `within_allowed_assignment?` guard. Added specs for the `case/in` and one-line `in` forms.

-----------------

* [ ] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [ ] Wrote [good commit messages][1].
* [X] Added tests.

### DIFF
--- a/changelog/fix_an_incorrect_autocorrect_for_performance_constant_regexp.md
+++ b/changelog/fix_an_incorrect_autocorrect_for_performance_constant_regexp.md
@@ -1,0 +1,1 @@
+* [#438](https://github.com/rubocop/rubocop-performance/issues/438): Fix an incorrect autocorrect for `Performance/ConstantRegexp` when a regexp is used as a pattern in pattern matching. ([@pcbeingused333][])
--- a/lib/rubocop/cop/performance/constant_regexp.rb
+++ b/lib/rubocop/cop/performance/constant_regexp.rb
@@ -43,7 +43,8 @@ module RuboCop
         end
 
         def on_regexp(node)
-          return if within_allowed_assignment?(node) || !include_interpolated_const?(node) || node.single_interpolation?
+          return if within_allowed_assignment?(node) || within_pattern_matching?(node) ||
+                    !include_interpolated_const?(node) || node.single_interpolation?
 
           add_offense(node) do |corrector|
             corrector.insert_after(node, 'o')
@@ -54,6 +55,20 @@ module RuboCop
 
         def within_allowed_assignment?(node)
           node.each_ancestor(:casgn, :or_asgn).any?
+        end
+
+        # `/o` is invalid when a regexp is used as a pattern in pattern matching,
+        # so an offense should not be registered in that case.
+        def within_pattern_matching?(node)
+          pattern_matching = node.each_ancestor(:in_pattern, :any_match_pattern).first
+          return false unless pattern_matching
+
+          pattern = if pattern_matching.in_pattern_type?
+                      pattern_matching.children.first
+                    else
+                      pattern_matching.children[1]
+                    end
+          node.equal?(pattern) || node.each_ancestor.any? { |ancestor| ancestor.equal?(pattern) }
         end
 
         def_node_matcher :regexp_escape?, <<~PATTERN
--- a/spec/rubocop/cop/performance/constant_regexp_spec.rb
+++ b/spec/rubocop/cop/performance/constant_regexp_spec.rb
@@ -63,4 +63,19 @@ RSpec.describe RuboCop::Cop::Performance::ConstantRegexp, :config do
       str.match?(/\#{CONST}\#{do_something(1)}/)
     RUBY
   end
+
+  it 'does not register an offense when regexp is used as a pattern in `case`/`in`' do
+    expect_no_offenses(<<~RUBY)
+      case string
+      in /\#{CONST}/
+        do_something
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when regexp is used as a one-line `in` pattern' do
+    expect_no_offenses(<<~RUBY)
+      string in /\#{CONST}/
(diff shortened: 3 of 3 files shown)


pcbeingused333:
Rebased onto current `master` (post-1.27.0) and re-pushed. Worth flagging why this PR has never shown a check: the three runs from Jul 30 came back as `startup_failure` with **zero jobs executed** — CI never ran on this branch at all. The rebase re-triggered them and they now sit in `action_required`, so a maintainer clicking *Approve and run workflows* is what it takes from here.

Since there is nothing green to look at yet, here is what I verified locally on the rebased branch (`rake spec` all green, `rake internal_investigation` clean across 126 files, rubocop 1.90.0 from master).

**Why the offense has to go rather than the autocorrect getting fixed.** `/o` on an interpolated regexp inside a pattern is not a parse error, it is a compile error — `ruby -c` reports `Syntax OK` and the file then dies at iseq build with `NODE_IN: unknown node (NODE_ONCE)`, which is what @morissetcl hit in #438. There is no correction the cop can emit that keeps `/o` there, so on that path the offense itself is unactionable.

**Where `/o` actually breaks, against what the cop now does** (checked with `RubyVM::InstructionSequence.compile` on ruby 3.2.3):

| construct | `/o` compiles | offenses after this PR |
|---|---|---|
| `in /#{CONST}/` | ✗ `NODE_IN` | 0 |
| `in [/#{CONST}/, *]` | ✗ `NODE_IN` | 0 |
| `in [*, /#{CONST}/, *]` | ✗ `NODE_IN` | 0 |
| `in {a: /#{CONST}/}` | ✗ `NODE_IN` | 0 |
| `in /#{CONST}/ \| String` | ✗ `NODE_IN` | 0 |
| `x => /#{CONST}/` | ✗ `NODE_IN` | 0 |
| `x in /#{CONST}/` | ✗ `NODE_IN` | 0 |
| `in String if x.match?(/#{CONST}/)` | ✓ | 1 |
| regexp in the `in` branch body | ✓ | 1 |
| regexp in the `case` subject | ✓ | 1 |
| nested `case`/`in` inside a branch body | ✓ | 1 |

The guard only covers the pattern slot, so the cop keeps firing everywhere `/o` is still legal — including inside the body of an `in` branch, which is the case that would be easy to over-suppress by walking ancestors alone.

I have extended the spec to cover that whole table rather than just the two cases the first version had: array, find, hash and alternative patterns, the rightward `=>` form (gated `:ruby30`, since `TargetRubyVersion` defaults to 2.7 here), plus two positive specs asserting the offense and its autocorrect still fire in an `in` guard and in a branch body. All six "does not register" specs fail with the guard reverted; the two positive ones pass either way by design, which is what makes them useful as a fence against over-suppression.

Happy to reshape any of this — in particular if you would rather keep the offense and mark it non-correctable on that path, so the "extract it into a constant" advice survives inside pattern matching. I went the other way because the message leads with `/o` and an offense with no available correction is mostly noise, but it is your call.

pcbeingused333:
Following up. Since the rebase the workflows were approved and all 15 checks are green — they had never executed at all before, so this is the first time there's been anything to look at. The spec now covers the full table above, with the six negative cases verified red against a reverted guard.

Still happy to go the other way and keep the offense while marking it non-correctable on the pattern-matching path, if that's the direction you'd rather take. Just let me know.

PR #542, opened 2026-09-08
[Fix #536] Fix a false positive for `Performance/ReverseEach` when the value is used

### LINKED ISSUE #536: ReverseEach autocorrect changes implicit method return value
## Expected behavior

`Performance/ReverseEach` should not autocorrect an expression whose value is implicitly returned. PR #221 added return-value protection, but implicit method returns are still missed.

## Reproduction

```ruby
def original
  [1, 2, 3].reverse.each { |_item| }
end

def corrected
  [1, 2, 3].reverse_each { |_item| }
end

original  # => [3, 2, 1]
corrected # => [1, 2, 3]
```

The cop reports and safe-autocorrects the original expression.

## Suggested direction

Treat the send—or its parent block—as unsafe when `value_used?` is true, in addition to the existing explicit assignment/send/return checks. An isolated candidate with that guard passes all 9,299 specs on Ruby 4.0.6 and Ruby 3.2.11.

This appears to be an incomplete edge case of #36 / #221.

pcbeingused333:
Fixes #536.

### Problem

`reverse.each` returns the reversed receiver; `reverse_each` returns the original one. So the rewrite is only safe when the value is thrown away, and the cop already tries to protect that with `use_return_value?`:

```ruby
def use_return_value?(node)
  !!node.ancestors.detect do |ancestor|
    ancestor.assignment? || ancestor.send_type? || ancestor.return_type?
  end
end
```

That covers an assignment, a method chain and an explicit `return`, but not an **implicit method return**:

```ruby
def process
  [1, 2, 3].reverse.each { |e| puts e }   # flagged and safe-autocorrected
end

process # => [3, 2, 1] before, [1, 2, 3] after
```

### Fix

`use_return_value?` now asks `value_used?` on the `reverse.each` node (or its block). `value_used?` already knows about `begin`/`if`/`while`/array/interpolation/... contexts, so this also fixes the same class of miss for a used `if` branch etc., and a `reverse.each` that is genuinely a discarded statement is still flagged.

### Tests

Added specs for the implicit return (with and without a preceding statement), a used `if` branch, and a `reverse.each` followed by another statement (still an offense). The three new "does not register" specs fail on `master`.

`bundle exec rake spec`, `rake internal_investigation`, `rake documentation_syntax_check` and a `PARSER_ENGINE=parser_prism` run of the cop's spec are all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -632,3 +632,4 @@
 [@a-lavis]: https://github.com/a-lavis
 [@jbpextra]: https://github.com/jbpextra
 [@lovro-bikic]: https://github.com/lovro-bikic
+[@pcbeingused333]: https://github.com/pcbeingused333
--- a/changelog/fix_a_false_positive_for_performance_reverse_each.md
+++ b/changelog/fix_a_false_positive_for_performance_reverse_each.md
@@ -1,0 +1,1 @@
+* [#536](https://github.com/rubocop/rubocop-performance/issues/536): Fix a false positive for `Performance/ReverseEach` when the value of `reverse.each` is used, such as an implicit method return. ([@pcbeingused333][])
--- a/lib/rubocop/cop/performance/reverse_each.rb
+++ b/lib/rubocop/cop/performance/reverse_each.rb
@@ -46,9 +46,11 @@ module RuboCop
         private
 
         def use_return_value?(node)
-          !!node.ancestors.detect do |ancestor|
-            ancestor.assignment? || ancestor.send_type? || ancestor.return_type?
-          end
+          # `reverse.each` returns the reversed receiver while `reverse_each` returns the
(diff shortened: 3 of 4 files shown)


pcbeingused333:
The workflows are on `action_required` (fork gate) — could a maintainer press "Approve and run workflows"? Local verification so the approval isn't a leap of faith, on the current HEAD (up to date with `master`):

- `bundle exec rake documentation_syntax_check` — passes
- `bundle exec rspec spec/rubocop/cop/performance/reverse_each_spec.rb` — 24 examples, 0 failures
- `PARSER_ENGINE=parser_prism bundle exec rspec spec/rubocop/cop/performance/reverse_each_spec.rb` — green

The three new "does not register" specs (implicit method return ×2, used `if` branch) fail on `master` and pass with the change.

PR #543, opened 2026-09-08
[Fix #534] Fix a false positive for `Performance/RedundantBlockCall` on a shadowed block argument

### LINKED ISSUE #534: RedundantBlockCall misses shadowed block argument inside begin body
## Expected behavior

`Performance/RedundantBlockCall` should not replace a call to a block-local variable with `yield` when that name shadows the method block argument.

PR #220 added shadow protection, but it only checks when the method body itself is a block node. A preceding statement wraps the body in `begin` and bypasses the guard.

## Reproduction

```ruby
def original(&outer)
  marker = :begin
  [-> { :inner }].each do |outer|
    return [marker, outer.call]
  end
end

# autocorrect changes outer.call to yield
```

Before correction the method returns `[:begin, :inner]`; after correction it returns `[:begin, :outer]`.

## Suggested direction

Search nested block nodes in the complete body for a matching block argument instead of requiring `body.block_type?`. An isolated `each_node(:block)` candidate suppresses the unsafe correction and passes all 9,299 specs on Ruby 4.0.6 and Ruby 3.2.11.

This appears to be an incomplete edge case of #162 / #220.

pcbeingused333:
Fixes #534.

### Problem

`RedundantBlockCall` skips a method when its block parameter is shadowed by an inner block, but the guard only fires when the method body *is* a block node:

```ruby
def shadowed_block_argument?(body, block_argument_of_method_signature)
  return false unless body.block_type?
  ...
end
```

A single statement before the block wraps the body in `begin`, so `body.block_type?` is `false` and the shadow is missed:

```ruby
def original(&outer)
  marker = :begin
  [-> { :inner }].each do |outer|
    return [marker, outer.call]   # rewritten to `yield`
  end
end

# before: [:begin, :inner]   after autocorrect: [:begin, :outer]
```

### Fix

`shadowed_block_argument?` now scans every `block` node in the body for an argument that shadows the method's block parameter, instead of only checking the body node itself. This matches what #220 was meant to do. As before, the cop bails on the whole method when a shadow is found (conservative, no scope analysis).

### Tests

Added specs for a shadowing block preceded by a statement, and a nested shadowing block. Both fail on `master`. The existing counterpart specs — a non-shadowing `do |_block|`, and the direct `do |block|` with no preceding statement — still pass.

`bundle exec rake spec`, `rake internal_investigation`, and a `PARSER_ENGINE=parser_prism` run of the cop's spec are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -632,3 +632,4 @@
 [@a-lavis]: https://github.com/a-lavis
 [@jbpextra]: https://github.com/jbpextra
 [@lovro-bikic]: https://github.com/lovro-bikic
+[@pcbeingused333]: https://github.com/pcbeingused333
--- a/changelog/fix_a_false_positive_for_performance_redundant_block_call.md
+++ b/changelog/fix_a_false_positive_for_performance_redundant_block_call.md
@@ -1,0 +1,1 @@
+* [#534](https://github.com/rubocop/rubocop-performance/issues/534): Fix a false positive for `Performance/RedundantBlockCall` when a block argument is shadowed by a nested block and the method body is not itself a block. ([@pcbeingused333][])
--- a/lib/rubocop/cop/performance/redundant_block_call.rb
+++ b/lib/rubocop/cop/performance/redundant_block_call.rb
@@ -91,9 +91,13 @@ module RuboCop
         end
 
         def shadowed_block_argument?(body, block_argument_of_method_signature)
-          return false unless body.block_type?
+          return false unless body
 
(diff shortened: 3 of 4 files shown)


pcbeingused333:
The workflows are on `action_required` (fork gate) — could a maintainer press "Approve and run workflows"? Local verification on the current HEAD (up to date with `master`):

- `bundle exec rspec spec/rubocop/cop/performance/redundant_block_call_spec.rb` — green (and green under `PARSER_ENGINE=parser_prism`)

The two new "accepts" specs (shadow preceded by a statement, nested shadow) fail on `master`.

PR #544, opened 2026-09-09
[Fix #535] Require a pure receiver in `Performance/DoubleStartEndWith`

### LINKED ISSUE #535: DoubleStartEndWith unsafe autocorrect changes impure receiver evaluation
## Expected behavior

`Performance/DoubleStartEndWith` should not autocorrect when the shared receiver is impure, because combining the calls changes both evaluation count and result.

## Reproduction

```ruby
class Sequence
  attr_reader :calls

  def initialize
    @calls = 0
  end

  def next_string
    @calls += 1
    @calls == 1 ? "x" : "b"
  end
end

sequence = Sequence.new
sequence.next_string.start_with?("a") || sequence.next_string.start_with?("b")
# => true; receiver called twice
```

The safe-autocorrect result is:

```ruby
sequence.next_string.start_with?("a", "b")
# => false; receiver called once
```

## Suggested direction

Require `receiver&.pure?` in `#check`, alongside the existing argument purity check. This suppresses the unsafe offense in the model. Existing cop examples use a bare send receiver named `x`, so their fixtures would need to be changed to local-variable receivers where autocorrection remains valid.

Reproduced on Ruby 4.0.6 and Ruby 3.2.11. The targeted Songstats 4,195-file scan has no affected offense; this report is based on an isolated runtime model.

pcbeingused333:
Fixes #535.

### Problem

`str.start_with?(a) || str.start_with?(b)` is only equivalent to `str.start_with?(a, b)` when `str` is evaluated the same number of times. The cop already required the **arguments** of the second call to be pure:

```ruby
return unless receiver && second_call_args.all?(&:pure?)
```

but not the **receiver**, so it combined and autocorrected code with a side-effecting receiver:

```ruby
class Sequence
  def next_string
    @calls += 1
    @calls == 1 ? "x" : "b"
  end
end

sequence.next_string.start_with?("a") || sequence.next_string.start_with?("b")
# => true  (receiver called twice)

# autocorrected to:
sequence.next_string.start_with?("a", "b")
# => false (receiver called once)
```

### Fix

`check` now also requires `receiver.pure?`. A pure receiver (local variable, instance variable, constant, literal) is safe to reference once instead of twice; a bare method call or a method-call chain is not.

The existing specs used a bare `x` receiver, which parses as `(send nil :x)` — an impure method call — so they now use `@x`. Two regression specs cover a bare method-call receiver and a chained one.

`bundle exec rake spec`, `rake internal_investigation`, and a `PARSER_ENGINE=parser_prism` run are green; `project_spec.rb` (no `@safety`/config drift) passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -632,3 +632,4 @@
 [@a-lavis]: https://github.com/a-lavis
 [@jbpextra]: https://github.com/jbpextra
 [@lovro-bikic]: https://github.com/lovro-bikic
+[@pcbeingused333]: https://github.com/pcbeingused333
--- a/changelog/fix_double_start_end_with_impure_receiver.md
+++ b/changelog/fix_double_start_end_with_impure_receiver.md
@@ -1,0 +1,1 @@
+* [#535](https://github.com/rubocop/rubocop-performance/issues/535): Fix `Performance/DoubleStartEndWith` to not combine `start_with?`/`end_with?` calls when the shared receiver is not pure, since that changes how often the receiver is evaluated. ([@pcbeingused333][])
--- a/lib/rubocop/cop/performance/double_start_end_with.rb
+++ b/lib/rubocop/cop/performance/double_start_end_with.rb
@@ -10,6 +10,10 @@ module RuboCop
       # `IncludeActiveSupportAliases` configuration option is used to check for
       # `starts_with?` and `ends_with?`. These methods are defined by Active Support.
       #
+      # The receiver is only combined when it is pure (a local variable, an
+      # instance variable, a constant, ...), since the original code evaluates it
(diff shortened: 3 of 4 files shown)


pcbeingused333:
The workflows are on `action_required` (fork gate) — could a maintainer press "Approve and run workflows"? Local verification on HEAD (up to date with `master`):

- `bundle exec rspec spec/rubocop/cop/performance/double_start_end_with_spec.rb` — 38 examples, 0 failures (green under `PARSER_ENGINE=parser_prism` too)
- `bundle exec rspec spec/project_spec.rb` — 71 examples, 0 failures (no `@safety`/config/alphabetical drift)

The two new "impure receiver" specs (bare method call, method-call chain) fail on `master`.

